PR #6, opened 2025-04-15
Adds core types for flow diagram

tswistak:
Defines interfaces for core components such as nodes, edges, events, command handling, model adapters, and rendering.

These interfaces provide a type-safe foundation for building a flow diagram editor.

### DIFF
--- a/packages/core/src/types/edge.interface.ts
+++ b/packages/core/src/types/edge.interface.ts
@@ -1,0 +1,10 @@
+/**
+ * Interface representing an edge (connection) between nodes in the flow diagram
+ */
+export interface Edge {
+  id: string;
+  source: string; // source node id
+  target: string; // target node id
+  type?: string;
+  data?: Record<string, unknown>;
+} 
--- a/packages/core/src/types/node.interface.ts
+++ b/packages/core/src/types/node.interface.ts
@@ -1,0 +1,12 @@
+/**
+ * Interface representing a node in the flow diagram
+ */
+export interface Node {
+  id: string;
+  type: string;
+  position: {
+    x: number;
+    y: number;
+  };
+  data?: Record<string, unknown>;
+} 
--- a/packages/core/src/types/renderer.interface.ts
+++ b/packages/core/src/types/renderer.interface.ts
@@ -1,0 +1,26 @@
+import { Node } from './node.interface';
+import { Edge } from './edge.interface';
+
+/**
+ * Interface for viewport data needed for rendering
+ */
+export interface ViewportData {
+  scale: number;
+  position: {
+    x: number;
+    y: number;
+  };
+}
+
+/**
+ * Interface for the Renderer that handles drawing the flow diagram
+ */
+export interface Renderer {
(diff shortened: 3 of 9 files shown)


packages/core/src/types/renderer.interface.ts
```diff
@@ -0,0 +1,26 @@
...
+  position: {
+    x: number;
+    y: number;
+  };
+}
+
+/**
+ * Interface for the Renderer that handles drawing the flow diagram
+ */
+export interface Renderer {
```
dawid-aksamski, line 18: I'm not quite sure if we need Renderer interface definition. This implies that core/model will know about the existence of the renderer because it needs to call it's `draw` method (coupling).
Shouldn't the core/model emit model changed event once all data processing is finished?

packages/core/src/types/edge.interface.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Interface representing an edge (connection) between nodes in the flow diagram
+ */
+export interface Edge {
+  id: string;
+  source: string; // source node id
```
dawid-aksamski, line 6: Why not name these properties as what is stated in the comment? 😄

packages/core/src/types/node.interface.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Interface representing a node in the flow diagram
+ */
+export interface Node {
+  id: string;
+  type: string;
```
dawid-aksamski, line 6: In the `Edge` interface, the `type` field is optional, any reason behind it's mandatory here or optional there? 😛

packages/core/src/types/renderer.interface.ts
piotrblaszczyk: I think it can stay as is for now. Because:

combineLatest([
  store.nodes$,
  store.edges$,
  store.viewport$
]).subscribe(([nodes, edges, viewport]) => {
  renderer.draw(nodes, edges, viewport);
});

tswistak: As we discussed earlier:
- We don't want to make the renderer hardly dependent on the model. Model should be interchangeable according to needs (SignalStore, RxStore, YStore etc. yet they all still use AngularRenderer).
- Core should be aware of the renderer. Why shouldn't it be?
- Core will be responsible for virtualization. If so, it needs to control rendering.

> Shouldn't the core/model emit model changed event once all data processing is finished?

- It would imply, that we will need another, external message queue, what we haven't planned.

packages/core/src/types/edge.interface.ts
tswistak: Both Edge and Node are sample interfaces, I didn't pay much attention to how they are implemented right now 😄. I could as well do `type Edge = unknown` for now.

packages/core/src/types/node.interface.ts
tswistak: As above, it's a temporary implementation.

PR #8, opened 2025-04-16
Implements command interpreter for event handling

tswistak:
Adds a core command interpreter for managing system events.

This implementation provides functionality for registering, emitting, and unregistering event callbacks. It enables decoupling of components by allowing them to subscribe to specific event types and react accordingly, improving the modularity and maintainability of the system.

Also, adds .idea/ to .gitignore.

### DIFF
--- a/packages/core/src/command-interpreter.ts
+++ b/packages/core/src/command-interpreter.ts
@@ -1,0 +1,51 @@
+import type { CommandInterpreter, SystemEvent, SystemEventCallback } from './types/command-interpreter.interface';
+
+/**
+ * Core implementation of CommandInterpreter interface
+ * Handles event emission and registration of callbacks for system events
+ */
+export class CoreCommandInterpreter implements CommandInterpreter {
+    private callbacks: Map<SystemEvent['type'], SystemEventCallback[]> = new Map();
+
+    /**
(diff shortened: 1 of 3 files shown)


packages/core/src/command-interpreter.ts
```diff
@@ -0,0 +1,51 @@
...
+     * @param eventType Type of event to listen for
+     * @param callback Function to be called when event occurs
+     * @returns Function to unregister the callback
+     */
+    register(eventType: SystemEvent['type'], callback: SystemEventCallback): () => void {
+        if (!this.callbacks.has(eventType)) {
+            this.callbacks.set(eventType, []);
+        }
+
+        const callbacks = this.callbacks.get(eventType) as SystemEventCallback[];
```
dawid-aksamski, line 34: Why check if the map has `eventType` entry above and then cast the return type of `get` method instead of checking the returned type?
```typescript
const callbacks = this.callbacks.get(eventType);
if (callbacks === undefined) {
  this.callbacks.set(eventType, [callback]);
} else {
    callbacks.push(callback);
}
```

tswistak: You're right, I haven't noticed it after Cursor generated it 😅. However, it still works the same

PR #21, opened 2025-04-22
Setup angular-adapter package

dawid-aksamski:


packages/angular-adapter/tsconfig.json
```diff
@@ -0,0 +1,29 @@
+{
+  "compileOnSave": false,
+  "compilerOptions": {
+    "paths": {
+      "angular-adapter": ["./dist/angular-adapter"],
+      "angular-adapter-2": ["./dist/angular-adapter-2"]
```
MateuszDro: What is angular-adapter-2?

dawid-aksamski: I was testing something and later removed this project 😄
I'll remove this entry, thanks for pointing this out! 😄

PR #25, opened 2025-04-24
Implement logger middleware

mateusz-heblewski:


packages/logger-middleware/src/logger.middleware.ts
```diff
@@ -0,0 +1,28 @@
...
+      console.log('[AngularFlow] Action:', {
+        type: context.modelAction.name,
+        data: context.modelAction.data,
+      });
+    }
+
+    if (logState) {
+      console.log('[AngularFlow] State:', {
+        before: context.initialState,
+        after: stateDiff,
```
MateuszDro: It's actually not after but more of change that the previous middleware applied

PR #28, opened 2025-04-25
Handling model updates

MateuszDro:


packages/core/src/types/edge.interface.ts
```diff
@@ -5,7 +5,7 @@ export interface Edge {
   id: string;
   source: string; // source node id
   target: string; // target node id
-  selected: boolean;
+  data: Record<string, unknown>;
+  selected?: boolean;
```
dawid-aksamski, line 9: Why selected is now optional? 😛

packages/core/src/types/node.interface.ts
```diff
@@ -3,11 +3,11 @@
...
-  type: string;
-  selected: boolean;
   position: {
     x: number;
     y: number;
   };
-  data?: Record<string, unknown>;
+  data: Record<string, unknown>;
+  type: string;
+  selected?: boolean;
```
dawid-aksamski, line 12: Same as above comment 😊

packages/core/src/types/edge.interface.ts
MateuszDro: We will add many properties like this e.g. dragging, selectable, draggable and so on and they should be optional since we dont want to declare these properties every time when we are creating node or if user will create such. We should more likely handle situations in which these properties are undefined

PR #36, opened 2025-04-29
Add create/update/delete commands

MateuszDro:


packages/input-event-handler/src/actions/select.ts
```diff
@@ -1,11 +1,13 @@
-import { ActionWithPredicate } from '@angularflow/core';
+import { ActionWithPredicate, PointerEvent } from '@angularflow/core';
 
 export const selectAction: ActionWithPredicate = {
-  action: ({ target }, inputEventHandler) => {
-    if (!target) {
+  action: (event, inputEventHandler) => {
+    const pointerEvent = event as PointerEvent;
+
+    if (!pointerEvent.target) {
```
dawid-aksamski: Why not introduce a type guard to check what kind of event we got? 😊

packages/core/src/commands/add-update-delete.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export interface UpdateNodeCommand {
+  name: 'updateNode';
+  id: string;
+  node: Partial<Node>;
+}
+
+export const updateNode = (commandHandler: CommandHandler, command: UpdateNodeCommand): void => {
+  const { nodes } = commandHandler.flowCore.getState();
+  commandHandler.flowCore.applyUpdate(
```
dawid-aksamski, line 23: Should this be possible to update the `id` of a `Node`? 😛

packages/core/src/commands/add-update-delete.ts
```diff
@@ -0,0 +1,71 @@
...
+export interface UpdateEdgeCommand {
+  name: 'updateEdge';
+  id: string;
+  edge: Partial<Edge>;
+}
+
+export const updateEdge = (commandHandler: CommandHandler, command: UpdateEdgeCommand): void => {
+  const { edges } = commandHandler.flowCore.getState();
+  commandHandler.flowCore.applyUpdate(
+    { edges: edges.map((edge) => (edge.id === command.id ? { ...edge, ...command.edge, id: command.id } : edge)) },
```
dawid-aksamski: Same as above comment 😛

packages/core/src/commands/add-update-delete.ts
MateuszDro: It's not possible since always the id is passed as last parameter of the object and it accepts the id passed with command so the node we update. Same for edge 😀

packages/input-event-handler/src/actions/select.ts
MateuszDro: Will do 👍

packages/core/src/commands/add-update-delete.ts
dawid-aksamski: Exactly, then why we set the `id` explicitly, it can't change 😄

PR #52, opened 2025-05-08
Add linking mechanism

MateuszDro:


packages/angular-adapter/projects/angular-adapter/src/lib/components/diagram/angular-adapter-diagram.component.ts
```diff
@@ -65,7 +65,11 @@ export class AngularAdapterDiagramComponent {
   edgeTemplateMap = input<EdgeTemplateMap>(new Map());
 
   nodes = this.renderer.nodes;
-  edges = this.renderer.edges;
+  edges = computed(() => {
```
dawid-aksamski: Shouldn't the Core be responsible for managing the nodes, edges and viewport data (e.g. virtualization)?

packages/angular-adapter/projects/angular-adapter/src/lib/services/renderer/renderer.service.ts
```diff
@@ -10,10 +10,12 @@ export class RendererService implements Renderer {
...
   });
+  temporaryEdge = signal<Edge | null>(null);
 
-  draw(nodes: Node[], edges: Edge[], viewport: Viewport): void {
+  draw(nodes: Node[], edges: Edge[], metadata: Metadata): void {
     this.nodes.set(nodes);
     this.edges.set(edges);
-    this.viewport.set(viewport);
+    this.viewport.set(metadata.viewport);
+    this.temporaryEdge.set(metadata.temporaryEdge ?? null);
```
dawid-aksamski: Why place temporary edges in the metadata? Isn't this just an edge with `temporary` set to true? 😄

packages/input-event-handler/src/actions/linking.ts
```diff
@@ -0,0 +1,41 @@
+import { ActionWithPredicate, isPointerDownEvent, isPointerMoveEvent, isPointerUpEvent } from '@angularflow/core';
+
+// TODO: This implementation adds linking on right button click. When port is added, double click should be removed.
```
dawid-aksamski: Shouldn't this say following:
```suggestion
// TODO: This implementation adds linking on right button click. When port is added, right click should be removed.
```

packages/input-event-handler/src/actions/__tests__/linking.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    it('should call startLinking if pointer down event and target is a node', () => {
+      linkingAction.action(
+        { ...mockedPointerEvent, type: 'pointerdown', target: { type: 'node', element: mockedNode }, button: 2 },
+        mockInputEventHandler,
+        environment
+      );
+      expect(mockCommandHandler.emit).toHaveBeenCalledWith('startLinking', { source: mockedNode.id });
+    });
+
+    it('should call startLinkingFromPosition if pointer down event and target is not a node', () => {
```
dawid-aksamski, line 95: Won't this be called also when the target is `Node`? There's no `else` statement 😄

packages/angular-adapter/projects/angular-adapter/src/lib/components/diagram/angular-adapter-diagram.component.ts
MateuszDro: You are right, I will move adding temporaryEdge to edges in core

packages/angular-adapter/projects/angular-adapter/src/lib/services/renderer/renderer.service.ts
MateuszDro: Yes and no 😅
It would be much easier to make it just another edge since it will be less logic in middlewares or when passing to adapter and rendering...but...temporary edge will start influencing edges array which means if any logic will relate on edges array or listen on it changes it will as well react on e.g. startLinking when in fact this specific edge does not exist yet.

PR #53, opened 2025-05-09
Move input-event-handler to core + add position transform utils

MateuszDro:


packages/core/src/input-event-handler/input-actions/select.ts
```diff
@@ -0,0 +1,13 @@
+import { isPointerDownEvent, type InputActionWithPredicate } from '../../types';
+
+export const selectAction: InputActionWithPredicate = {
+  action: (event, flowCore) => {
+    console.log(event);
```
dawid-aksamski: Unnecessary `console.log`? 😄

PR #59, opened 2025-05-15
Node size calculator and resizable node

mateusz-heblewski:


apps/angular-demo/src/app/app.component.spec.ts
```diff
@@ -1,21 +1,64 @@
...
+  AngularAdapterDiagramComponent,
+  FlowCoreProviderService,
+  INodeTemplate,
+  NodeTemplateMap,
+} from '@angularflow/angular-adapter';
 import { loggerMiddleware } from '@angularflow/logger-middleware';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
+import { SignalModelAdapter } from '@angularflow/angular-signals-model';
+import { ModelAdapter } from '../../../../packages/core/dist/src/types';
```
dawid-aksamski: This import doesn't look quite right 😄

packages/core/src/input-event-handler/input-actions/resize.ts
```diff
@@ -0,0 +1,25 @@
...
+    if (node.sizeControlled) {
+      return;
+    }
+
+    const { width: newWidth, height: newHeight } = event;
+
+    flowCore.commandHandler.emit('resizeNode', {
+      id: node.id,
+      size: {
+        width: newWidth,
```
dawid-aksamski: Is there a point of aliasing these names? 😛

packages/core/src/input-event-handler/input-actions/__tests__/resize.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    it('should return false for resize events with wrong target', () => {
+      expect(resizeAction.predicate({ ...mockEvent, target: { type: 'edge', element: mockEdge } }, mockFlowCore)).toBe(
+        false
+      );
+    });
+  });
+
+  describe('action', () => {
+    it('should emit resizeNode if node', () => {
```
dawid-aksamski: `(...) if target is node`? 😄

packages/angular-adapter/projects/angular-adapter/src/lib/directives/node-size/node-size.directive.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+    // do nothing
+  }
+}
+
+describe('NodeSizeDirective', () => {
+  let directive: NodeSizeDirective;
+  let fixture: ComponentFixture<TestComponent>;
+  let eventMapperService: EventMapperService;
+
+  @Component({
```
dawid-aksamski: Can we define the test component outside of the `describe` scope? 😊

packages/angular-adapter/projects/angular-adapter/src/lib/directives/node-size/node-size.directive.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+    template: `<div
+      angularAdapterNodeSize
+      [sizeControlled]="sizeControlled"
+      [size]="size"
+      [eventTarget]="eventTarget"
+    ></div>`,
+    standalone: true,
```
dawid-aksamski: In Angular 19 `standalone` components are default. We don't need to define this explicitly 😊

packages/angular-adapter/projects/angular-adapter/src/lib/directives/node-size/node-size.directive.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+    eventMapperService = TestBed.inject(EventMapperService);
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(directive).toBeTruthy();
+  });
+
+  it('should set size when sizeControlled is true and size is provided', () => {
+    const spy = vi.spyOn(directive as unknown as { setSize: (width: number, height: number) => void }, 'setSize');
```
dawid-aksamski: Why are we hackcessing `private` methods of a directive instead of observing their result (in multiple places)? 😮

packages/angular-adapter/projects/angular-adapter/src/lib/components/node/angular-adapter-node.component.spec.ts
```diff
@@ -1,24 +1,55 @@
...
   PointerEnterEventListenerDirective,
   PointerLeaveEventListenerDirective,
   PointerUpEventListenerDirective,
 } from '../../directives';
+import { NodePositionDirective } from '../../directives/node-position/node-position.directive';
+import { NodeSelectedDirective } from '../../directives/node-selected/node-selected.directive';
+import { NodeSizeDirective } from '../../directives/node-size/node-size.directive';
 import { AngularAdapterNodeComponent } from './angular-adapter-node.component';
 
+@Directive({
```
dawid-aksamski, line 16: Is it necessary to mock this directive to test if it's on the component? 😛

apps/angular-demo/src/app/node-template/resizable-node/resizable-node.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  selector: 'app-resizable-node',
+  imports: [FormsModule, DecimalPipe],
+  templateUrl: './resizable-node.component.html',
+  styleUrls: ['./resizable-node.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class ResizableNodeComponent implements INodeTemplate {
+  private readonly flowCoreProvider = inject(FlowCoreProviderService);
+
+  text = model<string>('');
```
dawid-aksamski, line 16: What's this text for? 😄

apps/angular-demo/src/app/node-template/resizable-node/resizable-node.component.html
```diff
@@ -0,0 +1,39 @@
...
+        [checked]="data().sizeControlled"
+        (change)="onSizeControlChange($event)"
+        (keydown)="$event.stopPropagation()"
+        (keypress)="$event.stopPropagation()"
+        (keyup)="$event.stopPropagation()"
+      />
+      Control size
+    </label>
+  </div>
+  <input
```
dawid-aksamski, line 29: What's this input doing? 😄

apps/angular-demo/src/app/app.component.spec.ts
```diff
@@ -1,21 +1,64 @@
...
 import { AppComponent } from './app.component';
 
+@Component({
```
dawid-aksamski, line 10: What's the point of mocking this? 😛

packages/angular-adapter/projects/angular-adapter/src/lib/components/node/angular-adapter-node.component.spec.ts
mateusz-heblewski: Yes, because if we don't do it, the test will depend on the internal dependencies of the directive, and it contains a ResizeObserver, which we would need to mock. In my opinion, we shouldn't mock things inside the directive or rely on its implementation in the test, but rather mock the entire directive.

apps/angular-demo/src/app/node-template/resizable-node/resizable-node.component.ts
mateusz-heblewski: I reused things from text-node in resizable-node, because it is a good example to show how sizeControlled works. When we type text into the input, the node's size expands, displaying the entered text at the bottom — it adjusts to the content. When the sizeControlled flag is set, the node's size should not adjust to the content but should instead be controlled externally. This functionality makes it easy to demonstrate.

apps/angular-demo/src/app/node-template/resizable-node/resizable-node.component.html
mateusz-heblewski: Same answer as for the comment above 😄

apps/angular-demo/src/app/node-template/resizable-node/resizable-node.component.ts
dawid-aksamski: Makes sense 😄

packages/angular-adapter/projects/angular-adapter/src/lib/components/node/angular-adapter-node.component.spec.ts
dawid-aksamski: Got it, thanks for the explanation! 😄

apps/angular-demo/src/app/app.component.spec.ts
mateusz-heblewski: To ensure that the test is not dependent on the internal implementation of AngularAdapterDiagramComponent or any of its contents. Internally, it uses a directive that relies on ResizeObserver, which would need to be mocked. 😄

apps/angular-demo/src/app/app.component.spec.ts
mateusz-heblewski: You're right, and in fact it was not necessary, so i removed it 😄

packages/core/src/input-event-handler/input-actions/resize.ts
mateusz-heblewski: In my opinion no, but this code created by @MateuszDro, so let's wait for him 😄

packages/core/src/input-event-handler/input-actions/__tests__/resize.test.ts
mateusz-heblewski: Yes, my bad 😄

packages/angular-adapter/projects/angular-adapter/src/lib/directives/node-size/node-size.directive.spec.ts
mateusz-heblewski: Yep, good catch, moved outside 😄

packages/angular-adapter/projects/angular-adapter/src/lib/directives/node-size/node-size.directive.spec.ts
mateusz-heblewski: Right! Removed 😄

packages/angular-adapter/projects/angular-adapter/src/lib/directives/node-size/node-size.directive.spec.ts
mateusz-heblewski: These tests were generated by AI 😂 I forgot to get back to this 😄

PR #67, opened 2025-05-20
[Angular Adapter] Elements ordering & some directives simplification

dawid-aksamski:


packages/core/src/types/edge.interface.ts
```diff
@@ -62,4 +62,8 @@ export interface Edge {
    * Whether the edge is temporary.
    */
   temporary?: boolean;
+  /**
+   * The z-index of the edge.
+   */
+  zIndex?: number;
```
MateuszDro: I would call it order or zOrder since the z-index is more DOM / CSS specific

dawid-aksamski: I agree, I initially did that but changed it to `zIndex` due to the name of the task 😅 
Thanks!

PR #69, opened 2025-05-21
Add bringToFront and sendToBack commands

dawid-aksamski:


packages/core/src/command-handler/commands/z-order.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const changeZOrder = (commandHandler: CommandHandler, command: BringToFrontCommand | SendToBackCommand): void => {
+  const { nodes, edges } = commandHandler.flowCore.getState();
+  const { nodeIds, edgeIds } = getCommandTarget(commandHandler, command);
+  if (nodeIds.length === 0 && edgeIds.length === 0) {
+    return;
+  }
+
+  const zOrder = getZOrder(nodes, edges, command);
+  const newNodes = nodes.map((node) => (nodeIds?.includes(node.id) ? { ...node, zOrder } : node));
```
MateuszDro: Please add statement if nodeIds are undefined or empty array it should be just previous nodes, no need to map through all array + the array will still have same reference since we are not modyfing anything there 😉

packages/core/src/command-handler/commands/z-order.ts
```diff
@@ -0,0 +1,60 @@
+import { CommandHandler, Edge, Node } from '../../types';
+
+function getZOrder(nodes: Node[], edges: Edge[], command: BringToFrontCommand | SendToBackCommand): number {
+  if (command.name === 'bringToFront') {
+    return Math.max(...nodes.map((node) => node.zOrder ?? 0), ...edges.map((edge) => edge.zOrder ?? 0)) + 1;
+  }
+
+  return Math.min(...nodes.map((node) => node.zOrder ?? 0), ...edges.map((edge) => edge.zOrder ?? 0)) - 1;
```
MateuszDro, line 8: This one is tricky since we easly can get indexes as minus values which could be something that we don't want to, but maybe we want, hard to say 😅

packages/core/src/command-handler/commands/z-order.ts
dawid-aksamski: But I'm doing what you wrote when checking the result of `getCommandTarget` 😄

packages/core/src/command-handler/commands/z-order.ts
dawid-aksamski: Who said they can't be negative? 😄

MateuszDro: Probably no one xd

packages/core/src/command-handler/commands/z-order.ts
MateuszDro: Yes but only if both are empty, what if someone passes only one edge? We still iterate through nodes :D

dawid-aksamski: Okay, now I see what you mean 😛

PR #70, opened 2025-05-22
Add resize adornment in angular adapter

dawid-aksamski:
image

### DIFF
--- a/packages/core/src/types/node.interface.ts
+++ b/packages/core/src/types/node.interface.ts
@@ -48,15 +48,6 @@ export interface Node {
    * Whether the node is resizable.
    */
   resizable?: boolean;
-  /**
-   * Configurable properties of the resize adornment.
-   */
-  resizeAdornment?: {
-    handleSize?: number;
-    strokeWidth?: number;
-    color?: string;
-    backgroundColor?: string;
-  };
 }
 
 /**
--- a/packages/angular-adapter/projects/angular-adapter/src/lib/components/node/angular-adapter-node.component.spec.ts
+++ b/packages/angular-adapter/projects/angular-adapter/src/lib/components/node/angular-adapter-node.component.spec.ts
@@ -1,6 +1,7 @@
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { beforeEach, describe, expect, it } from 'vitest';
 
+import { Component, input } from '@angular/core';
 import {
   NodePositionDirective,
   NodeSelectedDirective,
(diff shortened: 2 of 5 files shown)


packages/core/src/types/event.interface.ts
```diff
@@ -32,10 +32,32 @@ export interface DiagramTarget {
...
+  | 'bottom-left'
+  | 'left';
+
+/**
+ * Type for resize handle target
+ */
+export interface ResizeHandleTarget {
+  type: 'resize-handle';
+  position: ResizeHandlePosition;
+  element: Node;
```
MateuszDro, line 54: I am thinking if we could just pass as an element { nodeId: string; position: ResizeHandlePosition; } but I don't have any arguments why this should be better so for me it is fine 😀

packages/core/src/types/node.interface.ts
```diff
@@ -44,6 +44,19 @@ export interface Node {
...
    */
   ports?: Port[];
+  /**
+   * Whether the node is resizable.
+   */
+  resizable?: boolean;
+  /**
+   * Configurable properties of the resize adornment.
+   */
+  resizeAdornment?: {
```
MateuszDro: I am not a big fan of this approach to be honest since with such approach with every single node we will add we need to remember to pass set of properties so in our model there will be a lot of redudant data if we want to have all nodes to have some custom styling of resizing.

I think the styling of resize adornment should be more per node template not per node. I know that for now we want to have at least option to style our default but I think we should make these properties as a global config for flow-core and in the future it will work in way that if user want to have custom resize he will pass resize component to his node template and we will be able to user this to style/replace the default one 😀

packages/core/src/types/event.interface.ts
dawid-aksamski: Then I'll leave it as it is 😄

packages/core/src/types/node.interface.ts
dawid-aksamski: Got it, so I'm moving this to flow core as configurable properties with default values 😄

**EDIT:** Although I'm not sure about this after thinking about this for a while.
So we'd override default values in the flow core config right after init but how resize adornment will now about the changes? Since this info isn't going to be in the model, redrawing diagram won't do much.
We'd need to create something like `configChanged` event that the adornment would listen for 😛 

@MateuszDro how would you approach this?

MateuszDro: Sooo, maybe metadata.config? We can set it up anytime we want, we have it in model, and we don't need to config it per node 😀

dawid-aksamski: Yeah, we could do that 😄 I'll update the code then to reflect that 😄

dawid-aksamski: So I moved the resize adornment config to metadata. But there's 1 caveat, it is consumed once by the resize adornment. So if the user changes it later, the it won't be detected. Should we account for that at this stage too?
If the answer is yes, then I see 2 ways of doing it:
-  in resize adornment component we start listening for model changes
- we update renderer interface and push the metadata instead of only viewport and read the viewport from metadata and other properties if necessary

Let me know your thoughts on this 😄

PR #71, opened 2025-05-23
Implement spatial hash

MateuszDro:


packages/core/src/spatial-hash/spatial-hash.ts
```diff
@@ -0,0 +1,240 @@
+import { Node } from '../types';
+import { Rect, RectWithId } from '../types/utils';
+import { doesRectsIntersect, getRect } from './utils';
+
+// TODO: When set methods will be supported widely use them instead of these functions
+const setDifference = <T>(set1: Set<T>, set2: Set<T>): Set<T> => {
```
tswistak: You can always do a polyfill - try to add these functions to Set's prototype if they don't exist already, and then we're ready to use more performant built-in funcions

packages/core/src/spatial-hash/spatial-hash.ts
```diff
@@ -0,0 +1,240 @@
...
+          checkedRects.add(rect.id);
+          if (doesRectsIntersect(rect, range)) {
+            addToCollection(rect);
+          }
+        }
+      }
+    }
+  }
+
+  private isSameRect(rect1: Rect, rect2: Rect) {
```
tswistak: This could be in utils

packages/core/src/types/node.interface.ts
```diff
@@ -44,6 +44,10 @@ export interface Node {
    * The ports of the node.
    */
   ports?: Port[];
+  /**
+   * The angle of the node from 0 to 1.
```
dawid-aksamski: Why from 0 to 1? 😄 Wouldn't it be more intuitive for it to be from 0 to 360 and then `mod 360` to normalize it? 😄

PR #72, opened 2025-05-26
Add edge labels

MateuszDro:


packages/angular-adapter/projects/angular-adapter/src/lib/components/port/angular-adapter-port.component.ts
```diff
@@ -15,29 +27,60 @@ import { AngularAdapterNodeComponent } from '../node/angular-adapter-node.compon
   },
 })
 export class AngularAdapterPortComponent implements OnInit, OnDestroy {
+  @HostBinding('attr.data-port-id') get dataPortIdAttr() {
```
dawid-aksamski: Please add host bindings directly in `host` definition of the component (preferred way) 😄

packages/angular-adapter/projects/angular-adapter/src/lib/utils/get-point-on-path.ts
dawid-aksamski, line 1: Aren't we missing tests to this file? 😄

PR #80, opened 2025-05-29
Improve middlewares approach

MateuszDro:


packages/logger-middleware/src/logger.middleware.ts
```diff
@@ -2,13 +2,14 @@ import type { Middleware } from '@angularflow/core';
...
       initialState: context.initialState,
-      finalState: state,
-      historyUpdates: context.historyUpdates,
+      finalState: context.state,
+      history: context.history,
+      initialUpdate: context.initialUpdate,
     });
 
-    return state;
+    next();
```
dawid-aksamski, line 13: Shouldn't we test if the middleware is calling this function? This seems like a mandatory call 😄

packages/core/src/flow-core.ts
```diff
@@ -133,14 +134,14 @@ export class FlowCore {
...
    * @param modelActionType Type of model action to apply
    */
-  applyUpdate(state: Partial<FlowState>, modelActionType: ModelActionType): void {
-    const initialState = this.getState();
-    const updatedState = { ...initialState, ...state };
-    const finalState = this.middlewareManager.execute(initialState, updatedState, modelActionType);
-    this.setState(finalState);
+  async applyUpdate(stateUpdate: FlowStateUpdate, modelActionType: ModelActionType): Promise<void> {
+    const finalState = await this.middlewareManager.execute(this.getState(), stateUpdate, modelActionType);
+    if (finalState) {
```
dawid-aksamski, line 142: Shouldn't we test if there's no final state, then we don't set the state? 😄

PR #87, opened 2025-06-09
AF-42-rotate

EmilMalanczak:


packages/core/src/middleware-manager/middlewares/node-rotation-snap.ts
```diff
@@ -0,0 +1,41 @@
+import { snapAngle } from '../../input-event-handler/input-actions/rotate/snap-angle';
+import type { FlowStateUpdate, Middleware } from '../../types';
+
+const SNAP_ANGLE = 15;
```
EmilMalanczak, line 4: Consider using factory functions for middlewares instead of duplicating this logic. Rather than copy/pasting code to handle different SNAP values, we could accept a snap parameter. This would require updating all middlewares for consistency, but would eliminate code duplication and make the codebase more maintainable.

packages/angular-adapter/projects/angular-adapter/src/lib/components/node/rotate/node-rotate-adornment.component.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  readonly isRotating = signal(false);
+  readonly data = input.required<Node>();
+  readonly handleSize = signal(24);
+  readonly color = signal('#1e90ff');
+  readonly backgroundColor = signal('#fff');
+  readonly showAdornment = computed(() => !!this.data().selected || this.isRotating());
+
+  ngAfterViewInit() {}
```
piotrblaszczyk: it can be removed

PR #93, opened 2025-06-16
AF-54 Tree Layout 

paulinaCzekaj:


packages/core/src/types/utils.ts
piotrblaszczyk, line 1: I think that tree layout types should be placed in a dedicated filed

packages/angular-adapter/projects/angular-adapter/src/lib/components/diagram/angular-adapter-diagram.component.html
```diff
@@ -28,3 +28,4 @@
   }
   <angular-adapter-marker-arrow />
 </angular-adapter-canvas>
+<angular-adapter-toolbar />
```
MateuszDro: I am not sure about adding toolbar in adapter (maybe toolbar itself is not a bad idea but the content of it should not be defined in adapter since adapter is our library). 
Let's treat angular-adapter as reactflow library. Button for layout should be defined in the application 😊

packages/core/src/flow-core.ts
```diff
@@ -231,4 +231,14 @@ export class FlowCore {
   getNearestPortInRange(point: { x: number; y: number }, range: number): Port | null {
     return getNearestPortInRange(this, point, range);
   }
+
+  /**
+   * Sets the tree layout
+   */
+  async treeLayout() {
+    const finalState = await this.middlewareManager.execute(this.getState(), {}, 'treeLayout');
```
MateuszDro: Let's call here this.applyUpdate({}, 'treeLayout')

packages/angular-adapter/projects/angular-adapter/src/lib/services/layout/layout.service.ts
MateuszDro, line 1: I think this one is not necessary

packages/core/src/middleware-manager/middlewares/tree-layout/tree-layout.ts
```diff
@@ -0,0 +1,77 @@
...
+    });
+
+    const nodeList = Array.from(nodeMap.values());
+    for (const node of nodes) {
+      if (!node.position) {
+        continue;
+      }
+
+      const currentNode = nodeList.find((layoutNode) => layoutNode.id === node.id);
+      if (!currentNode || (node.position.x === currentNode.position.x && node.position.y === currentNode.position.y)) {
```
MateuszDro: Let's use isSamePoint utility to check if node.position and currentNode.position are equal 😁

apps/angular-demo/src/app/app.component.ts
```diff
@@ -78,7 +74,7 @@ export class AppComponent {
         id: '2',
         source: '2',
         target: '3',
-        data: {},
+        data: { label: 'Label 1' },
```
MateuszDro: What does it do?

packages/core/src/middleware-manager/middlewares/tree-layout/tree-layout.ts
```diff
@@ -0,0 +1,77 @@
...
+export const treeLayoutMiddleware: Middleware = {
+  name: 'tree-layout',
+  execute: (context, next) => {
+    const {
+      state: { edges, nodes, metadata },
+      modelActionType,
+    } = context;
+    const shouldTreeLayout = checkIfShouldTreeLayout(modelActionType);
+    const shouldAutoLayout = checkIfShouldAutoTreeLayout(context);
+    const config = metadata.layoutConfiguration?.tree || TREE_LAYOUT_DEFAULT_CONFIG;
```
MateuszDro: I am not sure about one thing, it looks like if user wants to configure layoutConfiguration he needs to either not change at all or if he decides to change at least one property he is forced to pass all of the confiuration. It would be nice to have possibility to change only one property so in the metadata configuration every property should be optional and in here we will do something like: 

`config = { ...TREE_LAYOUT_DEFAULT_CONFIG, ...(metadata.layoutConfiguration || {}) }`

apps/angular-demo/src/app/app.component.ts
paulinaCzekaj: I switched between branches and something from another task about the edge template was left over. 🤦‍♀️

PR #99, opened 2025-06-17
Add group sizing middlewares

EmilMalanczak:


packages/angular-adapter/projects/angular-adapter/src/lib/components/node/angular-adapter-node.component.ts
```diff
@@ -30,7 +31,22 @@ import { NodeRotateAdornmentComponent } from './rotate/node-rotate-adornment.com
...
 export class AngularAdapterNodeComponent {
+  private readonly portsService = inject(UpdatePortsService);
+  private readonly flowCore = inject(FlowCoreProviderService);
+
   data = input.required<Node>();
 
   readonly rotate = computed(() => (this.data().angle ? `rotate(${this.data().angle}deg)` : ''));
+  readonly id = computed(() => this.data().id);
+
+  constructor() {
```
EmilMalanczak, line 42: not sure if its the right way to call update on angle change

packages/core/src/command-handler/commands/group-highlight.ts
```diff
@@ -0,0 +1,47 @@
+import type { CommandHandler, Node } from '../../types';
+
+export interface GroupHighlightCommand {
+  name: 'groupHighlight';
+  groupId: Node['id'];
+}
+
+export const groupHighlight = async (
```
MateuszDro: I think it should be called highlightGroup
groupHighlight - zgrupuj zaznaczenie
highlightGroup - zaznacz grupę 
😅

PR #105, opened 2025-06-24
AF-79 layout   set layout orientation for sub graphs

paulinaCzekaj:
image

### DIFF
--- a/packages/core/src/flow-core.ts
+++ b/packages/core/src/flow-core.ts
@@ -235,7 +235,7 @@ export class FlowCore {
   /**
    * Sets the layout
    */
-  layout(layout: 'Tree' | 'None') {
+  layout(layout: 'Tree') {
     switch (layout) {
       case 'Tree':
         this.applyUpdate({}, 'treeLayout');


packages/core/src/types/node.interface.ts
```diff
@@ -56,6 +57,14 @@ export interface Node {
    * Whether the node is treated as a group node.
    */
   isGroup?: boolean;
+  /**
+   * Node layout direction: 0, 90, 180, 270.
+   */
+  layoutAngle?: LayoutAngleType;
```
MateuszDro: I am not sure about these properties since they are very specific for tree layout and if we add other layouts later these settings will grow and grow. I think such layout specific properties could be part of data so if user want to modify a node layout settings we let them know that to change for example layoutAlignment they should set data.layoutAlignment or data.treeLayout.layoutAlignment property or something like this. Or maybe adding some more generic property like layout configuration and it can contains settings depending on layouts we are exposing to users internally. Something probably to consider but for sure I think we should not add these properties flat like this because for elk we will get too much properties there 😁

packages/core/src/utils/tree-layout/orientation-tree-layout.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  // Set initial offset for children based on layout direction and sign
+  let childOffsetX =
+    offsetX + (!isVertical ? (sign === -1 ? -config.levelGap : sign * (parentWidth + config.levelGap)) : 0);
+  let childOffsetY =
+    offsetY + (isVertical ? (sign === -1 ? -config.levelGap : sign * (parentHeight + config.levelGap)) : 0);
+
+  // Special case: adjust offset if parent is sing = -1 and a direction changed
+  if (sign === -1 && grandparentAngle !== parentAngle) {
+    if (!isVertical) childOffsetY = offsetY;
```
MateuszDro: childOffsetY = isVertical ? offsetX : offsetY

packages/core/src/utils/tree-layout/orientation-tree-layout.ts
```diff
@@ -0,0 +1,193 @@
...
+    if (sign === -1 && isAngleVertical(grandparentAngle) === isVertical) {
+      if (!isVertical) {
+        childOffsetX -= width;
+      } else {
+        childOffsetY -= height;
+      }
+    }
+
+    // Recursively layout the child subtree
+    const childBounds = makeTreeLayout(child, config, childOffsetX, childOffsetY, parentAngle); // ⬅️ przekazujemy parentAngle jako grandparentAngle
```
MateuszDro: Polish comment 😊

packages/core/src/flow-core.ts
```diff
@@ -233,12 +233,14 @@ export class FlowCore {
...
-  async treeLayout() {
-    const finalState = await this.middlewareManager.execute(this.getState(), {}, 'treeLayout');
-    if (finalState) {
-      this.setState(finalState);
+   * Sets the layout
+   */
+  layout(layout: 'Tree' | 'None') {
+    switch (layout) {
+      case 'Tree':
+      default:
```
EmilMalanczak: I would go with defensive approach and throw an error as default when unknown layout will be passed (typescript might complain but on the runtime that might be hard to debug)

packages/core/src/utils/tree-layout/build-tree-structure.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    topGroupMap.set(id, topId);
+  }
+
+  return topGroupMap;
+};
+
+/**
+ * Zwraca top-level grupę dla danego nodeId.
```
EmilMalanczak: I would stick to english language in comments

packages/core/src/utils/tree-layout/build-tree-structure.ts
```diff
@@ -0,0 +1,134 @@
...
+  edges.map(edge => ({
+    ...edge,
+    source: getTopGroupId(edge.source, topGroupMap),
+    target: getTopGroupId(edge.target, topGroupMap),
+  }));
+
+export const buildGroupsHierarchy = (nodeMap: Map<string, TreeNode>): TreeNode[] => {
+  const result: TreeNode[] = [];
+
+  for (const node of nodeMap.values()) {
```
EmilMalanczak: is there a reason for making 3 loops on the same datasets instead of single one?

packages/core/src/types/tree-layout.interface.ts
```diff
@@ -1,14 +1,29 @@
...
-export interface TreeNode {
-  id: string;
+export interface TreeNode extends Pick<Node, 'id' | 'position' | 'size' | 'layoutAngle' | 'type' | 'groupId'> {
   children: TreeNode[];
-  size?: { width: number; height: number };
-  position: { x: number; y: number };
+  groupChildren?: TreeNode[];
 }
+
+export type Rect = { minX: number; maxX: number; minY: number; maxY: number };
```
EmilMalanczak: We already have `Rect` type which is of type 
```tsx
 {
    x: number;
    y: number;
    width: number;
    height: number;
 }
```

so that will cause naming confusion especially with auto-imports. 

I've merged the group related PR's which contains proper type with the same interface as the proposed one - its `Bounds` at `packages/core/src/types/utils.ts`

packages/core/src/utils/tree-layout/orientation-tree-layout.ts
paulinaCzekaj: I changed childOffsetX and childOffsetY to siblingOffset and parentChildOffset. I think it will be more logical, as it refers to vertical and horizontal alignment.

apps/angular-demo/src/app/app.component.ts
```diff
@@ -41,8 +42,26 @@ export class AppComponent {
...
-      { id: '4', type: 'group', position: { x: 100, y: 400 }, data: { title: 'Group 1' }, resizable: true, angle: 0 },
-      { id: '5', type: 'group', position: { x: 300, y: 400 }, data: { title: 'Group 2' }, resizable: true, angle: 0 },
+      {
+        id: '4',
+        type: 'group',
+        isGroup: true,
+        position: { x: 100, y: 400 },
+        data: { title: 'Group 1' },
+        resizable: true,
+        angle: 0,
```
MateuszDro: Angle and zOrder could be removed since they are optional and set to "default values" 😊

packages/core/src/types/node-data.interface.ts
```diff
@@ -0,0 +1,17 @@
+import { LayoutAlignmentType, LayoutAngleType } from './tree-layout.interface.ts';
+
+export type NodeData = {
```
MateuszDro: To be honest when I think of it now we are a bit taking from the user one of the property. I think this layoutConfiguration should be on the same level as other properties since this is "our" property comping from our library and keep NodeData as it was 😅

packages/angular-signals-model/src/signal-model-adapter.ts
```diff
@@ -1,12 +1,16 @@
...
 
 @Injectable()
 export class SignalModelAdapter implements ModelAdapter {
   // Internal state signals
   private nodes = signal<Node[]>([]);
   private edges = signal<Edge[]>([]);
-  private metadata = signal<Metadata>({ viewport: { x: 0, y: 0, scale: 1 } });
+  private metadata = signal<Metadata>({
+    viewport: { x: 0, y: 0, scale: 1 },
+    layoutConfiguration: { tree: TREE_LAYOUT_DEFAULT_CONFIG },
```
MateuszDro: Since layoutConfiguration is optional I wouldn't put it in here. The model package is one of the models and users will have possibility to create their own. We shouldn't force them to declare such metadata in every single model like zustand and so on (I think even viewport could be handle better but let's keep it for now 😅).

packages/core/src/flow-core.ts
```diff
@@ -233,12 +233,15 @@ export class FlowCore {
...
-    if (finalState) {
-      this.setState(finalState);
+   * Sets the layout
+   */
+  layout(layout: 'Tree' | 'None') {
+    switch (layout) {
+      case 'Tree':
+        this.applyUpdate({}, 'treeLayout');
+        break;
+      default:
```
EmilMalanczak, line 243: either add a 'None' case to handle or remove the 'Node' from the available union

PR #106, opened 2025-06-24
Make emitting commands asynchronous

MateuszDro:


packages/core/src/input-event-handler/input-actions/pointer-move-selection.ts
```diff
@@ -102,16 +102,12 @@ export const pointerMoveSelectionAction: InputActionWithPredicate = {
...
         }
 
         // That means a group has been highlighted, so we need to clear it
         if (updateData.some((node) => Boolean(node.groupId))) {
-          // TODO: Add batching updates - due to race condition this is not applied correctly
-          // the initial state for the update is not updated
-          setTimeout(() => {
-            flowCore.commandHandler.emit('highlightGroupClear');
-          }, 0);
+          flowCore.commandHandler.emit('highlightGroupClear');
```
EmilMalanczak, line 146: missing `async` or is it done by purpose?

MateuszDro: No need here 😁

PR #107, opened 2025-06-24
AF-61 orthogonal edge

paulinaCzekaj:


packages/angular-adapter/projects/angular-adapter/src/lib/components/diagram/angular-adapter-diagram.component.html
```diff
@@ -21,10 +21,12 @@
     >
       @if (getEdgeTemplate(edge.type)) {
         <ng-container *ngComponentOutlet="getEdgeTemplate(edge.type); inputs: { data: edge }" />
-      } @else if (edge.type === 'straight' || !edge.type) {
+      } @else if (edge.type === 'straight' || edge.type === 'orthogonal' || !edge.type) {
         <angular-adapter-edge-straight [data]="edge" />
```
MateuszDro: angular-adapter-edge-straight meant to be only for straight but when I see that actually different routing is just a different way of calculating path I think we can rename it or even move it to angular-adapter-edge

packages/angular-adapter/projects/angular-adapter/src/lib/types/edges.ts
```diff
@@ -0,0 +1,9 @@
+export interface Point {
```
MateuszDro: Let's reuse Point interface from core

packages/core/src/middleware-manager/middlewares/edges-routing.ts
```diff
@@ -60,7 +62,20 @@ export const edgesStraightRoutingMiddleware: Middleware = {
         if (!shouldRoute) {
           return;
         }
+
         const points = getPoints(edge, nodesMap);
+        const source = points[0];
```
MateuszDro: I think it would be nice to seperate logics of calculating points for straight and orthogonal since when we add some difeerent routing I am a bit worried that it will get messy here. There can be even some repeated code like const points = getPoints(edge, nodesMap) but I think it would be nicer later. As well then in orthogonal you can get at first all of the points so not the first and last ones firstly and the the middle ones 😊

packages/core/src/middleware-manager/middlewares/edges-routing/edges-routing.ts
```diff
@@ -60,11 +74,12 @@ export const edgesStraightRoutingMiddleware: Middleware = {
...
           return;
         }
-        const points = getPoints(edge, nodesMap);
+
+        const { sourcePoint, targetPoint, points } = getPoints(edge, nodesMap);
 
         if (
           edge.points?.length === points.length &&
-          edge.points?.every((point, index) => isSamePoint(point, points[index]))
+          edge.points?.every((point, index) => isSamePoint(point, points[index] as Point))
```
MateuszDro: I think as Point is not necessary, I checked it locally and it does not throw error so I assume it is something left from previous changes 😁

PR #108, opened 2025-06-24
Single resize observer instance

EmilMalanczak:


packages/angular-adapter/projects/angular-adapter/src/lib/components/port/angular-adapter-port.component.ts
```diff
@@ -41,10 +40,11 @@ export class AngularAdapterPortComponent implements OnInit, OnDestroy {
 
   lastSide = signal<Port['side'] | undefined>(undefined);
   lastType = signal<Port['type'] | undefined>(undefined);
+  private isInitialized = signal(false);
 
   constructor() {
     effect(() => {
-      if (this.lastSide() !== this.side()) {
+      if (this.isInitialized() && this.lastSide() !== this.side()) {
```
EmilMalanczak, line 47: this is tiny improvement preventing unnecessary command emit on the start when `lastType` or `lastSide` is `undefined`

PR #112, opened 2025-06-27
Add middlewares metadata

EmilMalanczak:


packages/angular-signals-model/src/signal-model-adapter.ts
```diff
@@ -40,15 +55,28 @@ export class SignalModelAdapter implements ModelAdapter {
...
     return this.metadata();
   }
 
-  setMetadata(next: Metadata | ((prev: Metadata) => Metadata)): void {
+  setMetadata(next: TMetadata | ((prev: TMetadata) => TMetadata)): void {
     this.metadata.update((prev) => (typeof next === 'function' ? next(prev) : next));
   }
 
-  onChange(callback: (data: { nodes: Node[]; edges: Edge[]; metadata: Metadata }) => void): void {
+  updateMiddlewareMetadata<TName extends keyof TMetadata['middlewaresMetadata']>(
```
MateuszDro: I am not sure about this one and the typing in this model. We need to remember that model instance is something we will deliver with packages like this signal-model-adapter but the idea is that user can create its own models which just need to meet our interface which was simply (getNodes/setNodes getEdges/setEdges getMetadata/setMetadata onChange) as simply as this. User can e.g. decide to create the zustand store. Enforicing him to write much more methods with much more generics types forces him to write a lot of boilerplate just to have simple store for our library.

What's more our idea is that user cannot modify anything in model strictly, it has to go through our application so if he wants to update metada he needs to call flowCore.commandHandler.emit('updateMetadata').

packages/core/src/types/metadata.interface.ts
```diff
@@ -31,12 +31,11 @@ export interface LayoutConfiguration {
...
  */
-export interface Metadata {
+export interface Metadata<TMiddlewaresMetadata = unknown> {
   viewport: Viewport;
   temporaryEdge?: Edge | null;
   nodeResizeAdornmentConfig?: NodeResizeAdornmentConfig;
   highlightedGroup?: Node['id'] | null;
   layoutConfiguration?: LayoutConfiguration;
-
-  [key: string]: unknown;
```
MateuszDro, line 41: The idea of metadata is something like ModelData in goJS but with some more important and internal data. Which means user can set in metadata anything they want related to model if they write their own middleware. If user will write his own middlewares he won't need to config them because he will set it locally in his app, we just need to add some possibilities for configuring middlewares created in our library.

I think we should approach it a bit different to be honest. We creating middlewares like edge-routing and so on so user has to write the smallest amount of boilerplate to actually start with something. But if they want to do their own changes they can e.g. unregister default middlewares. My idea is that if user want to keep current already implemented in core middleware he can do it and if he wants only change its config properties we could expose method to update config for our internal middlewares. Otherwise if user writes his own middleware he doesn't need to have special config object for it since the code is on his side so he can do whatever he wants with the properties in logic of this middleware.

Only concern in above approach is how to update registered middleware if some configs has changed on fly like in react in useState or something like that.

PR #117, opened 2025-07-01
AF-62 angular adapter edge templates

paulinaCzekaj:


apps/angular-demo/src/app/edge-template/custom-bezier-edge/custom-bezier-edge.component.html
```diff
@@ -0,0 +1,5 @@
+<angular-adapter-custom-edge
```
EmilMalanczak, line 1: @piotrblaszczyk maybe we can figure out a better name for this? That looks weird from the developer perspective that is supposed to work with such API

piotrblaszczyk: Yes, but let's leave it as is for now. I'll create a task to standardize naming across the entire library API.

PR #118, opened 2025-07-02
AF-82 move rotation handle to top right corner

paulinaCzekaj:
image

### DIFF
--- a/apps/angular-demo/src/app/app.component.spec.ts
+++ b/apps/angular-demo/src/app/app.component.spec.ts
@@ -56,7 +56,7 @@ describe('AppComponent', () => {
 
   it('should initialize with logger middleware', () => {
     const middlewares = component.middlewares();
-    expect(middlewares.length).toBe(1);
-    expect(middlewares[0].name).toBe('logger');
+    expect(middlewares.length).toBe(7);
+    expect(middlewares[6].name).toBe('logger');
   });
 });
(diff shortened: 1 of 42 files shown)


packages/angular-adapter/projects/angular-adapter/src/lib/components/node/rotate/handle/rotate-handle.component.ts
```diff
@@ -12,6 +15,19 @@ import { ChangeDetectionStrategy, Component, HostBinding, input, output } from '
 })
 export class RotateHandleComponent {
   isRotating = input.required<boolean>();
+  private readonly flowCoreProvider = inject(FlowCoreProviderService);
+
+  metadata = this.getMetadata();
+
+  computedOffset = computed(() => {
+    const current = this.metadata['rotateHandleOffset'] as RotateHandleConfiguration;
+    return {
```
piotrblaszczyk, line 25: It’s better to avoid using `null` in situations like this — optional properties should default to `undefined` when not set

```
return {
  top: current?.top ?? -20,
  right: current?.right ?? -36,
  bottom: current?.bottom,
  left: current?.left,
};

```

packages/angular-adapter/projects/angular-adapter/src/lib/components/node/rotate/handle/rotate-handle.component.ts
```diff
@@ -12,6 +15,19 @@ import { ChangeDetectionStrategy, Component, HostBinding, input, output } from '
...
+      top: current?.top ?? -20,
```
piotrblaszczyk: Also, please assign default value to const

PR #124, opened 2025-07-09
AF-84  implement proper z index

paulinaCzekaj:


packages/core/src/middleware-manager/middlewares/z-order/z-order.ts
```diff
@@ -0,0 +1,93 @@
...
+    const edgesToUpdate: FlowStateUpdate['edgesToUpdate'] = [];
+    let nodesWithZOrder: Node[] = [];
+    let edgesWithZOrder: Edge[] = [];
+
+    if (isInit) {
+      nodesWithZOrder = initializeZOrder(nodesMap);
+    }
+
+    if (shouldSnap) {
+      const selectedZOrder = metadata.selectedZOrder || DEFAULT_SELECTED_Z_ORDER;
```
EmilMalanczak: I would go with middlewares config and put this `DEFAULT_SELECTED_Z_ORDER` into defaultConfig

packages/core/src/middleware-manager/middlewares/z-order/utils/assignNodeZOrder.ts
```diff
@@ -0,0 +1,22 @@
+import { Node } from '../../../../types';
+
+export function assignNodeZOrder(
+  node: Node,
+  nodesMap: Map<string, Node>,
+  nodesWithZOrder: Node[],
+  currentZOrder: number
+) {
+  nodesWithZOrder.push({ ...node, zOrder: currentZOrder });
```
EmilMalanczak: I would rather avoid mutating like this, for recursion this is fine but for the first call we do also mutate `nodesWithZOrder` declared at `z-order.ts` file. Instead would be better to have  `nodesWithZOrder= assignNodeZOrder(node, nodesMap, baseZOrder);`

packages/core/src/middleware-manager/middlewares/z-order/z-order.ts
```diff
@@ -0,0 +1,114 @@
...
+    let edgesWithZOrder: Edge[] = [];
+
+    if (isInit) {
+      nodesWithZOrder = initializeZOrder(nodesMap);
+    }
+
+    if (shouldSnapNode) {
+      const selectedZOrder = metadata.selectedZOrder || DEFAULT_SELECTED_Z_ORDER;
+
+      ['selected', 'groupId'].forEach((prop) => {
```
EmilMalanczak: I think it would be better to split this into 2 separate loops for better readability

packages/core/src/middleware-manager/middlewares/z-order/z-order.ts
```diff
@@ -0,0 +1,114 @@
...
+      }
+      nodesToUpdate.push({ id: node.id, zOrder: node.zOrder });
+    }
+
+    let newTemporaryEdge: Edge | undefined = undefined;
+
+    if (isEdgeAdded) {
+      if (metadata?.temporaryEdge) {
+        const selectedZOrder = metadata.selectedZOrder || DEFAULT_SELECTED_Z_ORDER;
+        newTemporaryEdge = { ...metadata?.temporaryEdge, zOrder: selectedZOrder };
```
EmilMalanczak: I don't think so if this middleware should have responsibility for managing the temporary edge at all

packages/core/src/middleware-manager/middlewares/z-order/z-order.ts
```diff
@@ -0,0 +1,114 @@
...
+      }
+    }
+
+    if (shouldSnapEdge) {
+      const zOrderMap = new Map(nodesWithZOrder.map((n) => [n.id, n.zOrder ?? 0]));
+
+      for (const edgeId of helpers.getAffectedEdgeIds(['selected'])) {
+        const edge = edgesMap.get(edgeId);
+        if (!edge) continue;
+        edgesWithZOrder.push(edge.selected ? { ...edge, zOrder: 1000 } : assignEdgeZOrder(edge, zOrderMap, nodesMap));
```
EmilMalanczak: hardcoded value

packages/core/src/middleware-manager/middlewares/z-order/z-order.ts
```diff
@@ -0,0 +1,114 @@
...
+      }
+    } else edgesWithZOrder = assignEdgesZOrder(edges, nodesWithZOrder, nodesMap);
+
+    for (const edge of edgesWithZOrder) {
+      const currentEdge = edges.find((layoutNode) => layoutNode.id === edge.id);
```
EmilMalanczak: layoutNode - is that correct name?

PR #139, opened 2025-07-17
AF-92 Fixed bug with resizing

paulinaCzekaj:


packages/core/src/command-handler/commands/resize-node.ts
```diff
@@ -17,7 +17,10 @@ export async function resizeNode(commandHandler: CommandHandler, command: Resize
     throw new Error(`Node with id ${command.id} not found.`);
   }
 
-  if (node.size?.width === command.size.width && node.size?.height === command.size.height) {
+  if (
+    (node.size?.width === command.size.width && node.size?.height === command.size.height) ||
```
MateuszDro: By the way, there is a utily function `isSameSize` which does exactly same thing but it improves readability of such if statements :D

packages/core/src/middleware-manager/middlewares/node-position-snap.ts
```diff
@@ -17,39 +17,82 @@ export const nodePositionSnapMiddleware: Middleware<'node-position-snap', NodePo
...
+            width = snapNumber(Math.round(snapNumber(nodeWidth, snapConfig.x ?? 10)), snapConfig.x ?? 0);
+          }
+        }
 
-      const originalNode = flowCore.getNodeById(node.id);
+        if (prevHeight !== nodeHeight) {
+          if (movedY) {
+            height = Math.max(Math.round(originalNode.position.y + prevHeight) - snappedY, 0);
+          } else {
+            height = snapNumber(Math.round(snapNumber(nodeHeight, snapConfig.y ?? 10)), snapConfig.x ?? 0);
```
lukasz-jazwa: snapConfig.y zamiast "x"

PR #142, opened 2025-07-18
[AF-114] Fix get point on path for bezier curves

librowski-synergy:
- `IEdgeTemplate` -> `EdgeTemplate`
- Create `getPointOnPath` implementations for each `DefaultRouting` and include `routing` as parameter
- Minor refactoring


apps/angular-demo/src/app/app.component.ts
```diff
@@ -2,7 +2,7 @@ import { ChangeDetectionStrategy, Component, signal, Type } from '@angular/core'
 import {
   AngularAdapterDiagramComponent,
   EdgeTemplateMap,
-  IEdgeTemplate,
```
lukasz-jazwa, line 5: Dla spójności zmieńmy też INodeTemplate na NodeTemplate

PR #160, opened 2025-07-23
[AF-101] Support pasting nodes under cursor

kbcki:
**Info:**
There's a problem with offset when pasting nodes, it looks like this PR should resolve the problem here: https://github.com/synergycodes/angularflow/pull/155

### DIFF
--- a/packages/core/src/input-events/handlers/paste/paste.test.ts
+++ b/packages/core/src/input-events/handlers/paste/paste.test.ts
@@ -26,6 +26,7 @@ describe('PasteEventHandler', () => {
 
     mockFlowCore = {
       commandHandler: mockCommandHandler,
+      clientToFlowPosition: vi.fn((x) => x),
     } as unknown as FlowCore;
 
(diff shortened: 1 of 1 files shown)


packages/core/src/command-handler/commands/copy-paste.ts
```diff
@@ -11,60 +11,139 @@ export interface CopyCommand {
...
+  return copiedEdges.map((edge) => {
     const newEdgeId = crypto.randomUUID();
-    return {
+    const newEdge: Edge = {
       ...edge,
       id: newEdgeId,
       source: nodeIdMap.get(edge.source) || edge.source,
       target: nodeIdMap.get(edge.target) || edge.target,
+      // Keep original port IDs since they're hardcoded in templates
+      sourcePort: edge.sourcePort,
```
lukasz-jazwa: Te dwa w sumie są niepotrzebne bo `...edge` to załatwia

PR #218, opened 2025-08-20
Add layout integration example

librowski-synergy:


apps/docs/src/components/angular/examples/layout-integration/layout-buttons.component.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  async onCustomLayout() {
+    const { nodes: finalNodes, edges: finalEdges } = await performLayout(
+      this.nodes(),
+      this.edges()
+    );
+
+    for (const node of finalNodes) {
+      console.log(node.position.y);
```
lukasz-jazwa: Console log

PR #234, opened 2025-09-01
[AF-180] Create label docs

librowski-synergy:


apps/docs/src/content/docs/internals/edges/labels.mdx
```diff
@@ -1,4 +1,65 @@
...
 description: Understanding how labels on edges work in NgDiagram
+sidebar:
+  order: 3
 ---
+
+import { LinkCard } from '@astrojs/starlight/components';
+import CodeSnippet from '@components/code-snippet/code-snippet.astro';
+
+Labels are visual elements attached to edges that can display text, buttons, or any custom content. They automatically position themselves along the edge path and follow the edge as it moves or changes shape.
+
```
lukasz-jazwa, line 12: Here I'd add example with 2 default nodes connected with an edge with label - just to show at the beginning what we're writing about

apps/docs/src/content/docs/internals/edges/labels.mdx
```diff
@@ -1,4 +1,65 @@
...
+<CodeSnippet
+  relativePath="edges/custom-edges/labeled-edge/labeled-edge.component.ts"
+  range={[8, 33]}
+  mark={[{ range: '4-6' }]}
+/>
+
+## Label Positioning
+
+Based on the usage of the `BaseEdgeLabelComponent`, NgDiagram tracks the label's position and size, ensuring they remain aligned with the edge's path. The `positionOnEdge` property determines where the label appears along the edge path. It can be set to any value from `0` to `1`, where `0.5` is the center of the edge. Labels automatically follow the edge path, regardless of the used [routing algorithm](/internals/edges/routing).
```
lukasz-jazwa: The begining sounds like from LLM. I'd omit first sentence and start with the second one

apps/docs/src/content/docs/internals/edges/labels.mdx
```diff
@@ -1,4 +1,65 @@
...
+## Adding labels
+
+Labels can be added and modified dynamically. One way to achieve this is by using diagram's model. The `ModifiableLabelEdgeComponent` shows how to map such data into an actual label implementation.
+
+Click on an edge to select it in the example below. Then fill the label's name and click "Set Label".
+It'll modify selected edge's data to include a new `label` property.
+
+import AddingLabels from '@components/angular/edges/labels/adding-labels/adding-labels.astro';
+
+<AddingLabels />
```
lukasz-jazwa, line 40: Here I'd also add the code which sets label when we click on a button, so user knows how to set it programmatically

apps/docs/src/components/angular/edges/labels/adding-labels/modifiable-label-edge.component.ts
```diff
@@ -0,0 +1,46 @@
...
+  template: `<ng-diagram-base-edge
+    [edge]="edge()"
+    [stroke]="selected() ? 'rebeccapurple' : 'white'"
+  >
+    @if (label()) {
+      <ng-diagram-base-edge-label
+        [id]="label().id"
+        [positionOnEdge]="label().positionOnEdge"
+      >
+        {{ label().content }}
```
lukasz-jazwa, line 21: I think it would be good to add some background to the label so text is easier to read

apps/docs/src/components/angular/edges/labels/adding-labels/diagram.component.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    const updatedEdge = {
+      ...edgeToUpdate,
+      data: {
+        label: this.label,
+      },
+    };
+
+    this.model.setEdges(
```
lukasz-jazwa: Modyfing model directly isn't the best idea in ngDiagram, as it will skip all the logic like middlewares etc. To do it properly use `ngDiagramModelService.updateEdgeData`.

In public api we'll think how to communicate to the user that modyfing model directly is not the best idea

PR #255, opened 2025-09-09
[AF-199] remove from group with shift key

oskargrochowski-sc:


packages/core/src/command-handler/commands/remove-from-group.ts
```diff
@@ -0,0 +1,46 @@
+import type { CommandHandler, Node } from '../../types';
+
+export interface RemoveFromGroupCommand {
+  name: 'removeFromGroup';
+  groupId: Node['id'];
```
kamilEf: Chyba jets typ Group wiec mozna uzyc

PR #294, opened 2025-09-19
AF-225 Disable resize when node has angle different than 0

kamilEf:


packages/ng-diagram/projects/ng-diagram/src/lib/components/node/resize/ng-diagram-node-resize-adornment.component.ts
```diff
@@ -20,4 +20,5 @@ export class NgDiagramNodeResizeAdornmentComponent extends NodeContextGuardBase
 
   linePositions: LinePosition[] = ['top', 'right', 'bottom', 'left'];
   handlePositions: HandlePosition[] = ['top-left', 'top-right', 'bottom-left', 'bottom-right'];
+  readonly enabled = computed(() => !this.nodeData()?.angle);
```
piotrblaszczyk: Instead of introducing a new computed signal, consider extending the existing `showAdornment` signal

PR #350, opened 2025-10-14
[AF-257] Environment layer

EmilMalanczak:
An attempt to extract the environment related things into common place. At the end its an extend of the `detect-environment` functionality.

### DIFF
--- a/apps/docs/src/content/docs/api/Types/PointerInputEvent.md
+++ b/apps/docs/src/content/docs/api/Types/PointerInputEvent.md
@@ -193,7 +193,7 @@ Returns the object whose event listener's callback is currently being invoked.
 
 #### Inherited from
 
-[`PointerInputEvent`](/docs/api/types/pointerinputevent/).[`currentTarget`](/docs/api/types/pointerinputevent/#currenttarget)
+`PointerEvent.currentTarget`
 
 ***
 
@@ -491,7 +491,7 @@ Returns true if event was dispatched by the user agent, and false otherwise.
 
 #### Inherited from
 
-[`PointerInputEvent`](/docs/api/types/pointerinputevent/).[`srcElement`](/docs/api/types/pointerinputevent/#srcelement)
+`PointerEvent.srcElement`
 
 ***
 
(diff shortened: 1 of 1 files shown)


packages/ng-diagram/projects/ng-diagram/src/lib/services/environment-provider/environment-provider.service.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  private checkIfClient(): boolean {
+    return typeof window !== 'undefined' && typeof document !== 'undefined';
+  }
+
+  public now(): number {
+    return this.isClient ? performance.now() : Date.now();
+  }
+
+  public generateId(): string {
```
EmilMalanczak, line 59: This could be used as replacement for usages of crypto for `default-flow-config` and `internal-id-middleware`

packages/ng-diagram/projects/ng-diagram/src/lib/services/input-events/input-events-router.service.ts
```diff
@@ -1,25 +1,150 @@
...
     return {
-      modifiers: BrowserInputsHelpers.getModifiers(event),
-      id: this.generateEventId(),
-      timestamp: performance.now(),
+      modifiers: this.getModifiers(event),
+      id: this.environment.generateId(),
+      timestamp: this.environment.now(),
+    };
+  }
+
```
EmilMalanczak, line 34: Wondering about this change: I colocated the former `browser-input-helpers` into `InputEventsRouterService` since they depend on `EnvironmentProviderService` and are used wherever this service is consumed. Keeping them under `eventHelpers` (bound to preserve context) reduces imports and keeps usage cohesive. 

I’m not fully convinced this is the best long-term shape—do you think this improves clarity, or should leave it as it's been before?

piotrblaszczyk: Ok. After analysis, I think it’s fine to go with that. My only comments are:

1. **Rename `eventHelpers` → `eventGuards`.**
   The current name, `eventHelpers`, doesn’t clearly convey the purpose of these methods. They’re not general helpers — they act as pre-flight validation guards that determine whether an event should be processed before creating and emitting core events.

2. **Add JSDoc to the service class.**
   Include a brief JSDoc comment explaining that this service acts as an adapter between Angular-specific DOM events and the framework-agnostic `ng-diagram` core. This helps clarify why the service has two related responsibilities.

piotrblaszczyk (approved):

PR #397, opened 2025-10-28
AF-253 Grid background

kamilEf:


apps/angular-demo/src/app/app.component.ts
```diff
@@ -60,7 +60,8 @@ export class AppComponent {
       },
     },
     background: {
-      dotSize: 40,
+      default: 'grid',
+      gridSize: 100,
```
lukasz-jazwa: I have 2 things which are change requests to this task:

1. This gridSize parameter should mean size of the smallest cell in the gridBackground to be in sync with drag and resize Snapping

2. Currently in dragSnapping it's possible to specify different x and y. The grid should follow the same strategy, so instead of using number we need to use Point

apps/angular-demo/src/app/app.component.ts
```diff
@@ -60,7 +60,8 @@ export class AppComponent {
       },
     },
     background: {
-      dotSize: 40,
+      default: 'grid',
```
lukasz-jazwa: Let's change it so the user can pass background type directly in background component, without having to use component + set type in config. It will be more declarative.

smth like:
```
<ng-diagram-background type="dots" />
```

lukasz-jazwa: if the prop is implemented, let's remove this default config here

packages/ng-diagram/projects/ng-diagram/src/lib/components/background/default/grid/grid-background.component.html
```diff
@@ -0,0 +1,25 @@
...
+    patternUnits="userSpaceOnUse"
+  >
+    @for (line of gridLines(); track $index) {
+      <line
+        [attr.x1]="line.x1"
+        [attr.y1]="line.y1"
+        [attr.x2]="line.x2"
+        [attr.y2]="line.y2"
+        [attr.stroke]="line.major ? majorStroke : minorStroke"
+        [attr.stroke-width]="line.major ? 1 : 0.5"
```
lukasz-jazwa: It should be possible to change stroke, strokeWidth and opacity via css variables

packages/ng-diagram/projects/ng-diagram/src/lib/components/background/default/grid/grid-background.component.ts
```diff
@@ -0,0 +1,57 @@
...
+  styleUrl: './grid-background.component.scss',
+  imports: [NgFor, NgIf],
+})
+export class NgDiagramGridBackgroundComponent extends BackgroundPatternBase {
+  private readonly flowCore = inject(FlowCoreProviderService);
+
+  protected readonly backgroundPattern = viewChild<ElementRef<SVGPatternElement>>('backgroundPattern');
+
+  readonly majorLineEvery = 5;
+  readonly minorStroke = 'var(--ngd-background-line-minor, #c2c0c0ff)';
```
lukasz-jazwa: those variables aren't defined anywhere

packages/ng-diagram/projects/ng-diagram/src/lib/components/background/default/grid/grid-background.component.ts
```diff
@@ -0,0 +1,57 @@
...
+  templateUrl: './grid-background.component.html',
+  styleUrl: './grid-background.component.scss',
+  imports: [NgFor, NgIf],
+})
+export class NgDiagramGridBackgroundComponent extends BackgroundPatternBase {
+  private readonly flowCore = inject(FlowCoreProviderService);
+
+  protected readonly backgroundPattern = viewChild<ElementRef<SVGPatternElement>>('backgroundPattern');
+
+  readonly majorLineEvery = 5;
```
lukasz-jazwa: We need to make it configurable via config, with ability to differentiate between horizontal and vertical major lines, so it's possible to have major line horizontally every 5 cells and vertically every 10 cells

packages/ng-diagram/projects/ng-diagram/src/lib/styles/styles.css
```diff
@@ -53,8 +53,8 @@
   --ngd-default-edge-label-color: var(--ngd-txt-primary-default);
   --ngd-default-edge-label-border-color: var(--ngd-colors-gray-500);
 
-  /* change to 400 when ready*/
   --ngd-background-dot-color: var(--ngd-colors-gray-500);
+  --ngd-background-grid-color: var(--ngd-colors-gray-400);
```
lukasz-jazwa: This variable, and variables from tokens.css aren't used anywhere

packages/ng-diagram/projects/ng-diagram/src/lib/components/background/default/grid/grid-background.component.scss
```diff
@@ -0,0 +1,13 @@
+.grid-background {
+  position: absolute;
+  width: 100%;
+  height: 100%;
+  top: 0px;
+  left: 0px;
+  pointer-events: none;
+  background-color: var(--ngd-background-color, #ffffff);
```
lukasz-jazwa: We probably shouldn't set background color

lukasz-jazwa (approved):

PR #407, opened 2025-11-03
AF-334 Add measured bounds, improve collision detection for rotated node

Jacek-Synergy:


packages/ng-diagram/projects/ng-diagram/src/core/src/spatial-hash/collision-detection.ts
lukasz-jazwa, line 1: Create unit tests for this file

PR #415, opened 2025-11-04
Af 330 ng diagram service clean up

piotrblaszczyk:


packages/ng-diagram/projects/ng-diagram/src/lib/public-services/ng-diagram.service.ts
```diff
@@ -23,8 +23,15 @@ import { NgDiagramBaseService } from './ng-diagram-base.service';
...
- * const ready = this.ngDiagramService.isInitialized();
+ * // Check if diagram is initialized (reactive signal)
+ * effect(() => {
+ *   if (this.ngDiagramService.isInitialized()) {
+ *     console.log('Diagram ready!');
+ *   }
+ * });
+ *
+ * // Access reactive config
+ * const cellSize = this.ngDiagramService.config().cellSize;
```
lukasz-jazwa: `this.ngDiagramService.config().background.cellSize`

packages/ng-diagram/projects/ng-diagram/src/lib/public-services/ng-diagram.service.ts
```diff
@@ -23,8 +23,15 @@ import { NgDiagramBaseService } from './ng-diagram-base.service';
...
  *
  * // Update configuration
  * this.ngDiagramService.updateConfig({ cellSize: 20 });
```
lukasz-jazwa: This update need to be fixed too

PR #436, opened 2025-11-12
AF-277 Docs Tailwind CSS

kamilEf:


lukasz-jazwa:
There is no support for the dark mode

lukasz-jazwa:
It would be nice if this example showed a different type of node than the same as in custom node example — but for now it can stay as is, and we can involve a designer to create some other type of node here.

apps/docs/src/content/docs/examples/tailwind-styling.mdx
```diff
@@ -0,0 +1,17 @@
...
+
+import CodeViewer from '@components/code-viewer/code-viewer.astro';
+import TailwindStyling from '@examples/tailwind-styling/tailwind-styling.astro';
+
+This example demonstrates how to use Tailwind CSS in ngDiagram.
+
+<br />
+
+<TailwindStyling />
+{/* <CodeViewer dirName="examples/tailwind-styling" /> */}
```
lukasz-jazwa: We should see the code

PR #437, opened 2025-11-13
[AF-353] Adjust examples to landing page

Jacek-Synergy:


apps/docs/src/components/angular/examples/custom-edge/diagram.component.scss
```diff
@@ -1,8 +1,4 @@
 .diagram {
   display: flex;
-  height: 20rem;
-
-  --ngd-default-edge-stroke: #cccccc;
-  --ngd-default-edge-stroke-hover: gray;
-  --ngd-default-edge-stroke-selected: blue;
+  height: 30rem;
```
piotrblaszczyk, line 3: Maybe let’s create a global CSS variable for the height in all examples — or even a global class we can use to avoid repeating it.

apps/docs/src/components/angular/examples/custom-edge/labeled-edge.component.ts
```diff
@@ -12,12 +12,12 @@ const STROKE_WIDTH_SELECTED = 4;
 @Component({
   template: `<ng-diagram-base-edge
     [edge]="edge()"
-    stroke="orange"
+    stroke="#ECAF40"
```
piotrblaszczyk: Let’s move to CSS variables to avoid hardcoding colors in templates.

apps/docs/src/components/angular/examples/custom-edge/sinusoid-edge.component.ts
```diff
@@ -9,8 +9,8 @@ import {
 @Component({
   template: `<ng-diagram-base-edge
     [edge]="customEdge()"
-    stroke="rebeccapurple"
-    sourceArrowhead="ng-diagram-arrow"
+    stroke="#4C75F2"
```
piotrblaszczyk: here the same

apps/docs/src/components/angular/examples/custom-edge/diagram.component.ts
piotrblaszczyk, line 1: we need to regenerate images for all updated examples. Those images are used on the overview page.

kamilEf (approved):

apps/docs/src/components/angular/examples/custom-edge/diagram.component.scss
kamilEf: I will do that on my side ;)

PR #444, opened 2025-11-18
[AF-356] Fix shortcut capture, events, collision with inputs

Jacek-Synergy:


packages/ng-diagram/projects/ng-diagram/src/lib/directives/input-events/keyboard-inputs/keyboard-inputs.directive.ts
```diff
@@ -29,24 +46,24 @@ export class KeyboardInputsDirective {
...
-          };
+      const event = matchingAction && matchingAction.createEvent(shortcut, baseEvent);
 
       if (event) {
         this.inputEventsRouter.emit(event);
       }
     }
   }
+
+  private isInputFieldFocused(event: Event): boolean {
```
piotrblaszczyk, line 65: We have similar logic in `input-events-router`. I think we can stick with yours (it’s more general) and remove the `isDeleteKeyPressed` method from `input-events-router`.

PR #450, opened 2025-11-19
AF-359 Add missing links to the API in docs

kamilEf:


apps/docs/src/content/docs/guides/background.mdx
```diff
@@ -8,7 +8,7 @@ sidebar:
 import CodeViewer from '@components/code-viewer/code-viewer.astro';
 import Background from '../../../components/angular/background/background.astro';
 
-The background of your diagram canvas can be easily customized in ngDiagram.
+The background of your diagram canvas can be easily customized in ngDiagram.\
```
lukasz-jazwa, line 11: This slash looks like typo

kamilEf: nope ;) it's next line bracket because normally it does not split with one "enter" but with two, but then it adds extra space between.

PR #461, opened 2025-11-24
[AF-368] Cannot move node with arrows with large snap step value

kamilEf:


lukasz-jazwa:
I found 2 issues:
- I set snapping step to 100 x 100 and diagram scale to 0.227 - when moving a node with keyboard it hides beyond screen edge
- I set snapping for node '8' step to 10 x 10 and to node '10' step 100 x 100, select both nodes and use keyboard to move them, only node '8' moves

PR #464, opened 2025-11-25
[AF-370] Introduce Landing Page Diagram example

piotrblaszczyk:


apps/docs/src/components/angular/examples/landing-page-diagram/diagram/diagram.component.ts
```diff
@@ -0,0 +1,73 @@
...
+    <div class="not-content diagram">
+      <ng-diagram
+        data-no-pan="true"
+        [model]="model"
+        [config]="config"
+        [nodeTemplateMap]="nodeTemplateMap"
+      >
+        <ng-diagram-background />
+      </ng-diagram>
+      <div></div>
```
kamilEf: do we need this div?

piotrblaszczyk: ha! good catch. Absolutely no :) Added by accident

lukasz-jazwa (approved):

PR #468, opened 2025-11-30
AF-245 Ports with labels

kamilEf:


packages/ng-diagram/projects/ng-diagram/src/lib/components/port/ng-diagram-port.component.scss
```diff
@@ -1,51 +1,70 @@
...
-  width: var(--ngd-port-size);
-  height: var(--ngd-port-size);
-
   transform: translate(-50%, -50%);
+  z-index: 2;
+  cursor: crosshair;
 
-  background: var(--ngd-port-background-color);
-  border: var(--ngd-port-border-size) solid var(--ngd-port-border-color);
+  // Shared transition styles
```
lukasz-jazwa: Please remove comments generated by llm

packages/ng-diagram/projects/ng-diagram/src/lib/styles/styles.css
```diff
@@ -66,19 +66,23 @@
...
 .ng-diagram-port-hoverable .ng-diagram-port:hover,
+.ng-diagram-port-hoverable .ng-diagram-custom-port:hover .content-projection,
+.ng-diagram-port-hoverable-over-node:hover:not(:has(.ng-diagram-rotate-handle:hover)):not(:has([data-rotating]))
+  .ng-diagram-custom-port
+  .content-projection,
 .ng-diagram-port-hoverable-over-node:hover:not(:has(.ng-diagram-rotate-handle:hover)):not(:has([data-rotating]))
   .ng-diagram-port {
-  background-color: var(--ngd-port-background-color-hover);
-  border-color: var(--ngd-port-background-color-hover);
+  background-color: var(--ngd-port-background-color-hover) !important;
```
lukasz-jazwa: Let's try to omit !importants in the library

apps/angular-demo/src/app/node-template/chip-node/chip-node.component.scss
```diff
@@ -0,0 +1,27 @@
...
+  border-radius: 16px;
+
+  /* Non-applicable port CSS variables for custom styling */
+  // --ngd-port-size is not needed for custom styled ports - size is determined by content
+
+  /* Applicable port CSS variables for custom styling */
+  // --ngd-port-background-color: yellow;
+  --ngd-port-border-size: 0px;
+  // --ngd-port-border-color: red;
+  // --ngd-port-background-color-hover: #1565c0
```
lukasz-jazwa: semicolon at the end of this line

packages/ng-diagram/projects/ng-diagram/src/lib/components/port/ng-diagram-port.component.scss
```diff
@@ -1,51 +1,70 @@
...
+  --ngd-transition-timing: ease-in-out;
+  --ngd-transition-duration: 0.1s;
 
-  transition-property: background-color, border-color, outline, box-shadow;
-  transition-timing-function: ease-in-out;
-  transition-duration: 0.1s;
-  transform-origin: center;
+  // Styles for port with defined content
+  .content-projection:not(:empty) {
+    display: flex;
```
piotrblaszczyk, line 18: The styles below are identical for both "empty" and "custom" ports, so we can share them here to avoid duplication. For instance:

```css
  /* Shared port styles */
  .content-projection:not(:empty),
  &:has(.content-projection:empty) {
    display: flex;
    background: var(--ngd-port-background-color);
    border: var(--ngd-port-border-size) solid var(--ngd-port-border-color);

    transition-property: var(--ngd-transition);
    transition-timing-function: var(--ngd-transition-timing);
    transition-duration: var(--ngd-transition-duration);
    transform-origin: center;
  }
  ```
  
  But the question is. Do we want force an user to animate custom ports?

File: packages/ng-diagram/projects/ng-diagram/src/lib/components/port/ng-diagram-port.component.ts [hunk 5]
```diff
@@ -66,7 +72,11 @@ export class NgDiagramPortComponent extends NodeContextGuardBase implements OnIn
   side = input.required<Port['side']>();
 
   get portClass(): string {
-    return `ng-diagram-port ${this.side()}`;
+    if (!this.hasContent) {
```
piotrblaszczyk: we can simplify it a bit 

```typescript
get portClass(): string {
  return `${this.hasContent ? 'ng-diagram-custom-port' : 'ng-diagram-port'} ${this.side()}`;
}
```

packages/ng-diagram/projects/ng-diagram/src/lib/components/port/ng-diagram-port.component.scss
kamilEf: this was my comment :D but I will remove that :D

packages/ng-diagram/projects/ng-diagram/src/lib/components/port/ng-diagram-port.component.scss
kamilEf: Info: we agreed on refactoring these styles, adding a variable for border-radius, and fixing naming. Thanks ;)

apps/docs/src/components/angular/examples/custom-ports/diagram.component.ts
```diff
@@ -29,69 +31,99 @@ import { NodeComponent } from './node/node.component';
...
     zoom: {
       max: 3,
       zoomToFit: {
-        onInit: true,
-        padding: 100,
+        onInit: false,
       },
     },
+    edgeRouting: {
+      defaultRouting: 'orthogonal',
```
lukasz-jazwa: we don't have to set it, orthogonal is the default routing

packages/ng-diagram/projects/ng-diagram/src/core/src/types/edge.interface.ts
```diff
@@ -120,3 +120,18 @@ export interface EdgeLabel {
    */
   size?: Size;
 }
+
+/**
+ * The origin point options for port placement.
+ * @category Types/Model
+ */
+export type OriginPoint =
+  | 'leftTop'
```
lukasz-jazwa: In other libraries vertical position is used first in such names so let's use:
- topLeft
- centerLeft
- bottomLeft
- topCenter
- center
- bottomCenter
- topRight
- centerRight
- bottomRight

packages/ng-diagram/projects/ng-diagram/src/lib/components/port/ng-diagram-port.component.scss
```diff
@@ -1,53 +1,88 @@
...
+  width: fit-content;
+  height: fit-content;
   inset: 0;
-  border-radius: 50%;
-  cursor: crosshair;
-
   box-sizing: content-box;
-  width: var(--ngd-port-size);
-  height: var(--ngd-port-size);
+  z-index: 2;
```
lukasz-jazwa, line 7: Why 2 ? What content is overlapping the port without this setting ? It seems like a place which may be broken if ngDiagram assigns higher zIndex to the node

packages/ng-diagram/projects/ng-diagram/src/lib/components/port/ng-diagram-port.component.ts
```diff
@@ -65,8 +87,14 @@ export class NgDiagramPortComponent extends NodeContextGuardBase implements OnIn
    */
   side = input.required<Port['side']>();
 
+  /**
+   * The origin point for the port (e.g., leftTop, centerMiddle, rightBottom).
```
lukasz-jazwa: Let's add more explanation here, so user knows the purpose of this attribute without having to switch to documentation

apps/docs/src/components/angular/examples/custom-ports/socket-node/socket-node.component.html
```diff
@@ -0,0 +1,46 @@
...
+      top: 50%;
+      left: 50%;
+      transform: translate(-50%, -50%);
+      font-weight: bold;
+    "
+  >
+    {{ text() }}
+  </span>
+</div>
+<ng-diagram-port
```
lukasz-jazwa, line 17: On dark mode those ports have additional white border, let's remove it so it's consistent with light mode - I understand that on light mode there is the same white border but it's invisible there

apps/docs/src/content/docs/examples/custom-ports.mdx
```diff
@@ -22,24 +22,38 @@ This example demonstrates how to create custom ports in ngDiagram by customizing
 
 Two nodes are rendered, each with four ports (left, top, right, bottom).
 
+- Ports can contain any custom content, such as text, emoji, SVG, or images.
 - The left port of the second node uses a data property to apply red background.
```
lukasz-jazwa: We need to update this description, as it points to previous diagram

packages/ng-diagram/projects/ng-diagram/src/lib/components/port/ng-diagram-port.component.scss
kamilEf: It was here before, but I will check that

apps/docs/src/content/docs/examples/custom-ports.mdx
kamilEf: I removed that and added in section below something about custom content

lukasz-jazwa (approved):

PR #479, opened 2025-12-08
AF-365 Add grab cursor on background and investigate further cursor c…

kamilEf:
…hanges


packages/ng-diagram/projects/ng-diagram/src/lib/directives/input-events/panning/panning.directive.ts
```diff
@@ -109,4 +112,13 @@ export class PanningDirective implements OnDestroy {
       event.zoomingHandled
     );
   }
+
+  private toggleGrabbingCursor(isGrabbing: boolean): void {
```
piotrblaszczyk, line 116: we can simplify it:

`this.elementRef.nativeElement.classList.toggle('grabbing', isGrabbing);`

PR #487, opened 2025-12-09
[AF-386] Fix missing arrowheads safari

piotrblaszczyk:


apps/docs/src/components/angular/edges/arrowheads/custom-arrowhead/custom-arrowheads.component.ts
```diff
@@ -1,87 +1,109 @@
...
-            stroke="context-stroke"
-          />
-        </marker>
+    <!-- @mark-start -->
+    <!-- Square Arrowhead for Target -->
+    <ng-diagram-marker>
+      <svg>
+        <defs>
+          <marker
+            id="square-arrowhead-target"
```
kamilEf, line 14: If we want to use ID we should create a unique value for this. This might create some issues in future due to multiple DOM objects with the same ID. Please review all other ID's in this PR as well ;)

piotrblaszczyk: We have unique ids for all markers

PR #488, opened 2025-12-09
AF-376 Properties sidebar example - initial state out of sync

kamilEf:


lukasz-jazwa:
Just one thing - let's add some border to the label input, because right now it looks like a readonly field

PR #493, opened 2025-12-11
Af 383 transaction measurements

lukasz-jazwa:


packages/ng-diagram/projects/ng-diagram/src/core/src/flow-core.ts
```diff
@@ -223,28 +231,62 @@ export class FlowCore {
...
+      transactionOptions = options;
+      results = await this.transactionManager.transaction(nameOrCallback, callback, transactionOptions ?? {});
     }
 
     if (results.commandsCount > 0) {
+      if (transactionOptions?.waitForMeasurements) {
+        const internalOptions = transactionOptions as InternalTransactionOptions;
+        this.measurementTracker.trackStateUpdate(
+          results.results,
+          internalOptions._measurementDebounceTimeout,
```
lukasz-jazwa, line 278: Undocumented properties to customize timeouts for discovering measurements

packages/ng-diagram/projects/ng-diagram/src/core/src/transaction-manager/transaction-manager.ts
```diff
@@ -1,27 +1,43 @@
 import type { FlowCore } from '../flow-core';
 import type { FlowStateUpdate, LooseAutocomplete, ModelActionType, ModelActionTypes } from '../types';
-import type { TransactionCallback, TransactionResult } from '../types/transaction.interface';
+import type { TransactionCallback, TransactionOptions, TransactionResult } from '../types/transaction.interface';
 import { Transaction } from './transaction';
 
 export class TransactionManager<TFlowCore extends FlowCore = FlowCore> {
   private transactionStack: Transaction[] = [];
+  private pendingOptions: TransactionOptions | null = null;
```
lukasz-jazwa, line 8: Support for nested transactions - inner transactions will inherit options from outer transaction

packages/ng-diagram/projects/ng-diagram/src/core/src/types/transaction.interface.ts
```diff
@@ -1,5 +1,53 @@
...
+  waitForMeasurements?: boolean;
+}
+
+/**
+ * Internal options for measurement tracking. These are escape hatches for edge cases.
+ * Not part of the public API - use by casting to `any`.
+ *
+ * @internal
+ */
+export interface InternalTransactionOptions extends TransactionOptions {
```
lukasz-jazwa, line 44: Undocumented props to edit timeouts of measurement tracker, internal type to have it typed in core, but not exported via public api

packages/ng-diagram/projects/ng-diagram/src/lib/public-services/ng-diagram.service.ts
```diff
@@ -313,20 +315,87 @@ export class NgDiagramService extends NgDiagramBaseService {
...
+   * await this.ngDiagramService.transaction(() => {
+   *   this.ngDiagramModelService.addNodes([node1, node2]);
+   * }, { waitForMeasurements: true });
+   */
+  transaction(callback: () => void, options: TransactionOptions): Promise<TransactionResult>;
+  transaction(
+    callback: (() => void) | (() => Promise<void>),
+    options?: TransactionOptions
+  ): void | Promise<TransactionResult> {
+    let isAsync = false;
```
lukasz-jazwa, line 386: Some tricks to make this function backward compatible - if user passes sync callback without options - like previously, it will return void instead of Promise - like previously

PR #504, opened 2025-12-17
AF-390 Resizing group doesn't work correctly with rotated children

kamilEf:


packages/ng-diagram/projects/ng-diagram/src/core/src/utils/group-size.ts
```diff
@@ -32,7 +33,10 @@ export const calculateGroupBounds = (
         throw new Error(`calculateGroupBounds: child node ${node.id} does not have both width and height defined`);
       }
 
-      const nodeBounds = getBoundsFromRect(getRect(node));
+      // Use measuredBounds if available (accounts for rotation), otherwise compute it
```
lukasz-jazwa: Let's just remove this comment, as the code is self-explanatory

PR #510, opened 2025-12-29
AF-395 D&D HTML object (not palette node) to the diagram results in e…

kamilEf:
…rror

### DIFF
--- a/packages/ng-diagram/projects/ng-diagram/src/lib/directives/input-events/palette-drop/palette-drop.directive.ts
+++ b/packages/ng-diagram/projects/ng-diagram/src/lib/directives/input-events/palette-drop/palette-drop.directive.ts
@@ -52,14 +52,14 @@ export class PaletteDropDirective {
     const baseEvent = this.inputEventsRouterService.getBaseEvent(event);
     let parsedData = {};
 
-    if (!this.paletteService.draggedNode()) {
-      return console.error(PALETTE_DROP_NOT_NODE_ERROR());
-    }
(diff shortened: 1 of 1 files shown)


packages/ng-diagram/projects/ng-diagram/src/lib/directives/input-events/palette-drop/palette-drop.directive.ts
```diff
@@ -43,7 +43,7 @@ export class PaletteDropDirective {
       try {
         parsedData = JSON.parse(dataString);
       } catch (error) {
-        console.error(PALETTE_DROP_JSON_PARSE_ERROR(dataString, error));
+        return console.error(PALETTE_DROP_JSON_PARSE_ERROR(dataString, error));
```
lukasz-jazwa: In my opinion we should show warning instead of error

lukasz-jazwa: If it is possible to discover with 100% accuracy that we're currently dragging palette item and we're catching error here, we should show `console.error` otherwise we should show `console.warning`, or maybe nothing at all

kamilEf: I changed that implementation. Please check now ;)

lukasz-jazwa (approved):

PR #537, opened 2026-01-29
[AF-250] Introduce minimap component

piotrblaszczyk:


packages/ng-diagram/projects/ng-diagram/src/lib/components/minimap/ng-diagram-minimap.calculations.ts
```diff
@@ -0,0 +1,142 @@
...
+ * so individual nodes stay in their original diagram coordinates.
+ *
+ * @param node - The diagram node
+ * @param defaultSize - Optional default size for nodes without explicit size
+ */
+export const extractNodeBounds = (
+  node: Node,
+  defaultSize: { width: number; height: number } = DEFAULT_NODE_SIZE
+): MinimapBounds => {
+  const size = node.size ?? defaultSize;
```
lukasz-jazwa, line 103: What about `measuredBounds` ? This property probably is more general than size which is set by user -

piotrblaszczyk: `measuredBounds` returns the bounding box after rotation, which makes it larger than what I actually need in this case.

Additionally, it does not provide any information about the rotation itself, which I want to visualize on the minimap.

Regarding to missing size information - this is handled by ResizeObserver. When the node’s size changes due to its content, ResizeObserver updates the value accordingly, so all cases are covered.

PR #548, opened 2026-02-06
Add a Touch Devices section to the box-selection article

piotrblaszczyk:


apps/docs/src/content/docs/guides/box-selection.mdx
```diff
@@ -42,3 +42,9 @@ You can customize the appearance of the box selection rectangle by using predefi
 <br />
 
 <Styled />
+
+## Touch Devices
+
+On touch devices, box selection is triggered by a long press gesture instead of holding `Shift`. Press and hold on an empty area of the canvas, then drag to create a selection box.
```
lukasz-jazwa: It doesn't have to be empty area, wherever will work

PR #622, opened 2026-03-20
[NGD-114] Wrap initializeModel body with untracked() so it can be saf…

piotrblaszczyk:
…ely used

inside computed/effect/linkedSignal

### DIFF
--- a/packages/ng-diagram/projects/ng-diagram/src/lib/model/initialize-model.ts
+++ b/packages/ng-diagram/projects/ng-diagram/src/lib/model/initialize-model.ts
@@ -26,7 +26,17 @@ import { stripEdgeRuntimeProperties, stripNodeRuntimeProperties } from './strip-
  *
  * // With an explicit injector (outside injection context)
  * model = initializeModel({ nodes: [...], edges: [...] }, this.injector);
+ *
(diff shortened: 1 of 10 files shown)


packages/ng-diagram/projects/ng-diagram/src/lib/model/initialize-model.ts
```diff
@@ -36,9 +36,11 @@ import { stripEdgeRuntimeProperties, stripNodeRuntimeProperties } from './strip-
  * @category Utilities
  */
 export function initializeModel(model: Partial<Model> = {}, injector?: Injector): ModelAdapter {
```
lukasz-jazwa, line 48: I think we should perform the same fix for `initializeModelAdapter` too. Because it is also used as standalone function when registering new model adapters

piotrblaszczyk: I was thinking about it, but adding `untracked` there would hide potential issues in user's code, rather then surfacing. It's becouse `initializeModelAdapter` doesn't create adapter - it takes whatever user passes in. It should be user's responsibility to handle it if custom adapter creates any effects/signals internally

lukasz-jazwa (approved):

lukasz-jazwa (approved):

PR #624, opened 2026-03-23
[NGD-116] Fix palette drag preview clipping when ancestor has overflo…

piotrblaszczyk:
…w:hidden

### DIFF
--- a/packages/ng-diagram/projects/ng-diagram/src/lib/components/palette/item/ng-diagram-palette-item.component.ts
+++ b/packages/ng-diagram/projects/ng-diagram/src/lib/components/palette/item/ng-diagram-palette-item.component.ts
@@ -41,10 +41,7 @@ export class NgDiagramPaletteItemComponent {
 
   /** @internal */
   onDragStart(event: DragEvent) {
-    const previewHtmlElement = this.paletteItemPreviewComponent()?.preview();
-    if (previewHtmlElement && previewHtmlElement.nativeElement) {
-      event.dataTransfer?.setDragImage(previewHtmlElement.nativeElement, 0, 0);
-    }
+    this.setDragPreviewImage(event);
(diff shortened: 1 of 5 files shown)


packages/ng-diagram/projects/ng-diagram/src/lib/components/palette/item/ng-diagram-palette-item.component.ts
```diff
@@ -58,4 +55,27 @@ export class NgDiagramPaletteItemComponent {
...
+   */
+  private setDragPreviewImage(event: DragEvent) {
+    const previewHtmlElement = this.paletteItemPreviewComponent()?.preview();
+    if (!previewHtmlElement?.nativeElement || !event.dataTransfer) {
+      return;
+    }
+
+    const clone = previewHtmlElement.nativeElement.cloneNode(true) as HTMLElement;
+    clone.style.position = 'fixed';
+    clone.style.top = '0';
```
lukasz-jazwa: Maybe just to be sure there are no "flashes" we should position it at '-9999px' ?

lukasz-jazwa (approved):

packages/ng-diagram/projects/ng-diagram/src/lib/components/palette/item/ng-diagram-palette-item.component.ts
piotrblaszczyk: it inherits css from the cloning element, so by default we have  `left: -1000px;`

piotrblaszczyk: Fixed. Indeed, while the clone is only on-screen for a single frame (removed via requestAnimationFrame), it could theoretically be visible on non-Safari browsers. Good catch!

PR #646, opened 2026-04-15
feat: add built-in self loop edge routing support

adithyaakrishna:
Description:

Implements first-class self-loop support in ng-diagram

What changed:

- Added new built-in self-loop routing
- Added self-loop metadata in routing context (selfLoopIndex, selfLoopCount)
- Added edgeRouting.selfLoop config + defaults
- Updated edge routing flow to detect and route source === target edges correctly
- Updated linking validation to allow valid same-node self-connections
- Added/updated tests and demo self-loop examples

### DIFF
--- a/apps/angular-demo/src/app/data/default-model.ts
+++ b/apps/angular-demo/src/app/data/default-model.ts
@@ -264,5 +264,31 @@ export const defaultModel: DiagramModel = {
       targetPort: 'port-right-1',
       targetArrowhead: 'ng-diagram-arrow',
     },
+    {
+      id: '20',
+      source: '5',
+      target: '5',
+      data: { labelPosition: 0.5 },
+      type: 'labelled-edge',
+    },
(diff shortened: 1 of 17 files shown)


adithyaakrishna:
We use `ng-diagram` in prod and this feature is smth I built for our use-case, hence thought to contribute it to upstream as well :)

PS. Pls do lmk incase yall need any changes to be done, would be glad to do it

lukasz-jazwa:
Hello @adithyaakrishna , thank you for your feature idea. It's great to hear that you use ng-diagram in prod. 

Currently we're busy with 2 updates we're going to add to the library - bug fixes bundle + orgchart application template. Both will be available next week, and once we release them, we'll come back to your PR and evaluate it

adithyaakrishna:
@lukasz-jazwa sounds good, thansk for the update :)

PR #671, opened 2026-04-24
Batching all port and label changed props

lukasz-jazwa:


packages/ng-diagram/projects/ng-diagram/src/core/src/updater/internal-updater/internal-updater.ts
```diff
@@ -84,62 +87,90 @@ export class InternalUpdater implements Updater {
...
+      if (!measuredPort) {
+        return false;
+      }
 
-  private getPortsToUpdate(
-    node: Node,
-    ports: Pick<Port, 'id' | 'size' | 'position'>[]
-  ): Pick<Port, 'id' | 'size' | 'position'>[] {
-    const measuredPortsMap = this.buildMeasuredPortsMap(node);
+      if ('side' in portChanges && portChanges.side !== measuredPort.side) {
```
piotrblaszczyk: The check uses `'side' in portChanges`, so a payload like `{ side: undefined }` (e.g. from destructuring an optional Angular input) will pass the filter when `measuredPort.side` is set, because `undefined !== 'left'`. The update then propagates downstream and can overwrite a valid side with undefined. Same concern for `type`.

Suggestion: tighten to `portChanges.side !== undefined && portChanges.side !== measuredPort.side` (and analogously for type), or document that callers must omit the key entirely when they don't intend to change it.

packages/ng-diagram/projects/ng-diagram/src/core/src/updater/internal-updater/internal-updater.ts
```diff
@@ -84,62 +87,90 @@ export class InternalUpdater implements Updater {
...
+      const label = measuredLabelsMap.get(labelId);
+      if (!label) {
+        return false;
+      }
+
+      if ('positionOnEdge' in labelChanges && labelChanges.positionOnEdge !== label.positionOnEdge) {
+        return true;
+      }
+
+      if ('size' in labelChanges) {
```
piotrblaszczyk: When labelChanges.size is `undefined/null`, the filter always returns true and lets the update through. If clearing a label's size is not a legal operation, this becomes a source of redundant emits whenever a caller passes the key with an empty value. If it is legal, the comment above the function ("filters out updates where no property actually differs") is misleading - clearing an already-cleared size still passes.

Suggestion: either guard with `labelChanges.size !== undefined` and treat clearing as no-op when `label.size` is also missing, or clarify the intent in the doc comment.

piotrblaszczyk (approved):

PR #672, opened 2026-04-24
Batch processor race conditions fix

lukasz-jazwa:


packages/ng-diagram/projects/ng-diagram/src/core/src/batch-processor/batch-processor.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+  }
+
+  /**
+   * Flushes all pending operations.
+   * Executes sequentially: deletes → updates → adds.
+   * Each callback is awaited before the next to prevent read-modify-write races.
+   */
+  private async flush(): Promise<void> {
+    const allDeletes = this.takeAll(this.pendingDeletes);
```
piotrblaszczyk: The flush sequence is `deletes → updates → adds`. This produces a counter-intuitive result when an entity is added and then deleted within the same tick (e.g. dynamic Angular inputs, rapid reactive updates):

1. `deleteCallback` runs first against state that doesn't yet contain the new entity → no-op (or error).
2. `addCallback` then creates the entity.

**Result: the entity exists at the end of the tick even though the caller intended to delete it.**

The conventional ordering for batched mutation pipelines is `add → update → delete`, or merging intents per-id (cancel both sides if the same id has both add and delete). The current order seems optimized for the virtualization re-entry case, but at the cost of normal dynamic-port/dynamic-label scenarios.

**Tests to add in `batch-processor.spec.ts`:**

- `processAdd(key, X)` followed by `processDelete(key, X.id)` in the same tick → assert final state has `X` removed.
- `processAdd(key, X)` followed by `processUpdate(key, { id: X.id, ...changes })` in the same tick → assert update is applied to `X` (i.e. add must precede update).
- `processDelete(key, X.id)` followed by `processAdd(key, X)` in the same tick (re-add scenario) → assert `X` exists at the end. Document this as the intended virtualization path.

packages/ng-diagram/projects/ng-diagram/src/core/src/batch-processor/batch-processor.ts
```diff
@@ -0,0 +1,92 @@
...
+  private readonly pendingUpdates = new Map<string, TUpdate[]>();
+  private readonly pendingDeletes = new Map<string, string[]>();
+
+  private addCallback: FlushCallback<TAdd> | null = null;
+  private updateCallback: FlushCallback<TUpdate> | null = null;
+  private deleteCallback: FlushCallback<string> | null = null;
+
+  private flushScheduled = false;
+
+  processAdd(key: string, item: TAdd, onFlush: FlushCallback<TAdd>): void {
```
piotrblaszczyk, line 47: If two different callers invoke `processAdd` with different `onFlush` callbacks within the same tick, only the **last** callback runs — but it receives **all accumulated data** from both callers. Today this works because `DirectPortUpdateStrategy` and `VirtualizedPortUpdateStrategy` share a single `portBatchProcessor` instance and both emit the same bulk command, so the callbacks are semantically equivalent. But this is not enforced by types and is fragile: a future caller that wants a different emit target will silently have its data routed through someone else's callback.

Possible mitigations:

- Assert the callback identity (`if (this.addCallback && this.addCallback !== onFlush) throw ...`).
- Store the callback only on the first call per flush window (ignore subsequent ones if equal, throw if different).
- Make the callback per-key instead of global.

**Tests to add in `batch-processor.spec.ts`:**

- Two `processAdd` calls in the same tick with the same callback for different keys → assert callback fires once with both keys present in the map.
- Two `processAdd` calls in the same tick with **different** callbacks → assert/document the chosen behavior (currently: only the second callback fires with both keys' data; ideally: throw or run both).
- Same scenarios for `processUpdate` and `processDelete`.
- Cross-operation: `processAdd` and `processDelete` in the same tick → verify both callbacks are invoked with their respective data, and confirm neither callback overwrites the other.

packages/ng-diagram/projects/ng-diagram/src/core/src/updater/internal-updater/direct-port-update-strategy.ts
```diff
@@ -4,29 +4,28 @@ import type { Port } from '../../types';
 import type { PortUpdateStrategy } from './port-update-strategy.interface';
 
 /**
- * Direct port update strategy - uses per-node batching.
- * Used when virtualization is disabled.
+ * Direct port update strategy - used when virtualization is disabled.
  */
 export class DirectPortUpdateStrategy implements PortUpdateStrategy {
```
piotrblaszczyk: After this refactor, both strategies use the same `processAdd` / `processUpdate` / `processDelete` on the shared `portBatchProcessor` and emit the same `*Bulk` commands. The only meaningful difference is that `VirtualizedPortUpdateStrategy.addPort` filters out ports that are already present in `node.measuredPorts` (re-entry into viewport), while `DirectPortUpdateStrategy.addPort` does not.

`updatePorts` and `deletePort` are byte-for-byte copy-paste between the two classes. The header comments are also stale — `DirectPortUpdateStrategy` still says "uses per-node batching" even though per-node batching no longer exists.

This is a code smell that the Strategy abstraction may no longer be earning its weight. A simpler shape would be a single class with a `skipReAddIfMeasured: boolean` flag, or a small inline branch in `addPort`.

Probably out of scope for this PR, but worth either:

- Collapsing the two strategies into one and deleting the interface.
- Or leaving a `TODO` linking to a follow-up issue, and at minimum fixing the stale "per-node batching" comment in `direct-port-update-strategy.ts`.

No new tests required — existing strategy tests should still cover both code paths.

packages/ng-diagram/projects/ng-diagram/src/core/src/batch-processor/batch-processor.ts
lukasz-jazwa: resolved

packages/ng-diagram/projects/ng-diagram/src/core/src/batch-processor/batch-processor.ts
lukasz-jazwa: resolved  (Make the callback per-key instead of global)

packages/ng-diagram/projects/ng-diagram/src/core/src/updater/internal-updater/direct-port-update-strategy.ts
lukasz-jazwa: resolved - strategies removed, logic inlined as both strategies use the same bulk actions

piotrblaszczyk (approved):

PR #673, opened 2026-04-24
Batch size measurements

lukasz-jazwa:


packages/ng-diagram/projects/ng-diagram/src/lib/services/flow-resize-observer/flow-resize-processor.service.ts
```diff
@@ -152,9 +152,8 @@ export class FlowResizeBatchProcessorService {
    */
   private processNodeBatch(entries: ProcessedEntry[]): void {
```
piotrblaszczyk, line 158: The post-init path now emits `commandHandler.emit('updateNodes', { nodes: nodeSizeUpdates })` for **all** size updates - both nodes without an initial size and nodes whose size changed.

Before this PR, the size-changed branch went through `flowCore.updater.applyNodeSize(nodeId, size)`, which is implemented in `InternalUpdater` as:

```ts
applyNodeSize(nodeId, size): void {
  const node = this.flowCore.getNodeById(nodeId);
  if (!node || this.isResizing() || this.hasSameSize(node, size)) {
    return;
  }
  this.flowCore.commandHandler.emit('resizeNode', { id: nodeId, size });
}
```

`InternalUpdater.applyNodeSize` is unchanged in this PR - but the new call site **bypasses it entirely** for post-init size-changed nodes. As a result:

- The `isResizing()` early-return is no longer reached on this path. While the user is mid-drag on a resize handle, `actionStateManager.isResizing()` returns `true` and the resize action is the source of truth (with snapping, constraints, transforms). Previously, ResizeObserver-driven updates were suppressed during this window. Now they fire `updateNodes` and may compete with the resize action - risk of size flicker, fighting snapping, or overwriting transformed values with raw DOM measurements.
- The `isResizing` flag is still computed in `processNodeBatch`, but it now only gates the **port-measurement** follow-up - not the size update itself.

Either restore the guard (skip pushing to `nodeSizeUpdates` when `isResizing` is true), or document explicitly why ResizeObserver-driven updates are now expected to fire mid-resize.

packages/ng-diagram/projects/ng-diagram/src/lib/services/flow-resize-observer/flow-resize-processor.service.ts
```diff
@@ -165,23 +164,14 @@ export class FlowResizeBatchProcessorService {
...
-        continue;
-      }
-
-      if (!this.isSizeChanged(currentSize, size)) {
+      if (node.size && !this.isSizeChanged(node.size, size)) {
         continue;
       }
 
-      flowCore.updater.applyNodeSize(metadata.nodeId, size);
+      nodeSizeUpdates.push({ id: metadata.nodeId, size });
```
piotrblaszczyk, line 176: Size-changed nodes post-init are now routed through `'updateNodes'` instead of `'resizeNode'`, silently dropping resize-domain logic

To be clear about what changed: `commandHandler.emit('updateNodes', ...)` was already present in the post-init branch before this PR. The structural shape of `if (flowCore.isInitialized) { emit('updateNodes', ...) } else { ...applyNodeSize... }` is identical before and after. **The change is what gets pushed into the emitted array.**

Before this PR, the loop split entries into two paths:

- **Branch A** - no current size (`!currentSize`): pushed into `nodesNeedingInitialSize`, emitted as a single `updateNodes` post-loop.
- **Branch B** - size changed (`isSizeChanged(currentSize, size)`): called `flowCore.updater.applyNodeSize(id, size)` per-node *inside the loop*, which routes through `InternalUpdater.applyNodeSize` → emits `'resizeNode'` with the full domain pipeline.

After this PR, both branches are merged into a single `nodeSizeUpdates` array, and the post-loop `updateNodes` emit now carries entries from **Branch B** as well. So Branch A is unchanged, but **Branch B's command path silently switches from `'resizeNode'` to `'updateNodes'`**.

The two handlers are not equivalent:

- `resizeNode` (`resize-node.ts`) handles missing-initial-size via `handleMissingInitialSize`, applies `handleGroupNodeResize` (children bounds containment, `applyMinimumSizeConstraints`), and `handleSingleNodeResize` (minimum size, snapping).
- `updateNodes` (`add-update-delete.ts`) is a bare passthrough to `applyUpdate({ nodesToUpdate: nodes }, 'updateNodes')` - no constraints, no group containment, no snapping.

Concrete scenarios that regress post-init (only Branch B - size-changed nodes):

- A **selected** group node whose child reflows: ResizeObserver fires for the group → `updateNodes` is emitted → children bounds containment and minimum-size constraints are not applied.
- An **unselected** group node whose child reflows: this is the most surprising case. `resizeNode` early-returns at `resize-node.ts:120` for `isGroup(node) && !node.selected`, so previously the group's stored size was **not touched** during a child reflow - it was treated as user-set source of truth. Post-PR, `updateNodes` overwrites the stored size with the raw DOM measurement on every reflow. This silently mutates an invariant that previously held.
- A non-group node with dynamic content (e.g. content-driven CSS sizing) shrinks below `getMinNodeSize`: `updateNodes` accepts the raw size; previously `resizeNode` clamped it via `applyMinimumSizeConstraints`.
- Snapping configured for resize: previously applied via `handleSingleNodeResize` → `applySnappingIfNeeded`; now skipped on this path.

Branch A (no-initial-size) is unaffected: even pre-PR it bypassed `resizeNode`'s constraint pipeline, because `resizeNode`'s `handleMissingInitialSize` (`resize-node.ts:195-207`) itself just emits a bare `updateNode` without constraints. So the pre-PR `updateNodes` emit was effectively equivalent to what `resizeNode` would have done for that case. **The regression is exclusively about Branch B** - size-changed nodes that previously preserved the full resize-domain semantics and now do not.

Options:

- Keep the per-node `applyNodeSize` call for size-changed nodes (preserves `'resizeNode'` and its logic), and only batch the no-initial-size path (status quo before the PR).
- Or introduce a new bulk command like `'resizeNodesBulk'` that runs the same group/min-size/snapping pipeline as `'resizeNode'` but for an array of nodes, and emit that instead.
- Or explicitly document that ResizeObserver-driven post-init updates intentionally skip resize-domain logic, and accept the regression.

**Tests to add in `flow-resize-processor.service.spec.ts` and/or `resize-node.test.ts`:**

- **Selected** group node with children whose total bounds exceed a shrunk parent size → trigger ResizeObserver post-init → assert children-bounds containment is still applied.
- **Unselected** group node whose child reflows → trigger ResizeObserver post-init → assert the group's stored size is **not** silently overwritten with the raw DOM measurement (preserving the prior invariant), or explicitly document that this invariant is dropped.
- Non-group node with `getMinNodeSize` config → trigger ResizeObserver with smaller size post-init → assert size is clamped to minimum.
- Snapping enabled + ResizeObserver-driven update post-init → assert snapping is applied (or document drop).

**Suggested fix (combined with Comment 4):**

Re-introduce the pre-PR partition between Branch A and Branch B inside `processNodeBatch`:

- **Branch A - no current size** → batch via `commandHandler.emit('updateNodes', ...)` (preserves the original batching intent for palette drops / virtualization).
- **Branch B - size changed** → per-node `flowCore.updater.applyNodeSize(id, size)`, which already routes through `'resizeNode'` with the full pipeline (group containment, min-size, snapping) and respects the `isResizing()` guard.

This restores pre-PR semantics for size-changed nodes while keeping the bulk benefit for the case the batching was actually introduced for.

packages/ng-diagram/projects/ng-diagram/src/lib/services/flow-resize-observer/flow-resize-processor.service.ts
lukasz-jazwa: resolved - restored `isResizing` guards

packages/ng-diagram/projects/ng-diagram/src/lib/services/flow-resize-observer/flow-resize-processor.service.ts
lukasz-jazwa: We resolved this intentionally by choosing a different approach than any of the three options suggested. After analyzing
  resizeNode's domain logic, I concluded that none of it was appropriate for ResizeObserver-driven updates — it was
  user-resize logic that was historically applied to DOM measurements only because all sizes happened to route through
  applyNodeSize → resizeNode:

  - Min-size constraints / snapping: These are user-resize concerns. The DOM already rendered at its actual size — clamping
  or snapping a DOM measurement would fight the real layout.
  - Group containment: Prevents the user from dragging a group smaller than its children. Not relevant when the DOM is
  reporting the group's actual rendered size.
  - Unselected group skip (isGroup(node) && !node.selected): This means "you can't resize a group unless it's selected" — a
  user-interaction guard. For ResizeObserver, autoSize: false groups have explicit CSS dimensions, so the DOM reports the
  same size as stored → isSizeChanged already filters it as unchanged. autoSize: true groups should accept DOM measurements
  regardless of selection state.
  - isResizing guard: The only guard that actually matters for ResizeObserver. I restored it in applyNodeSizes — it
  suppresses size-changed nodes during active user drag, but allows initial sizes through (nodes without size yet need their
   first measurement regardless).

  What I did:

  1. processNodeBatch now collects all size updates and delegates to updater.applyNodeSizes() — no branching, no domain
  logic at the call site
  2. InternalUpdater.applyNodeSizes handles the guards (isResizing for size-changed nodes, isSameRect dedup) and emits
  'updateNodes'
  3. InternalUpdater.applyNodeSize (singular) also emits 'updateNodes' instead of 'resizeNode' for consistency
  4. resizeNode command is now exclusively for user-initiated resize (drag handles, public API)

  Addressing each scenario:

  - Selected group, child reflows: Previously resizeNode applied containment. Now updateNodes accepts the DOM size. Correct
  — the DOM is the source of truth for content-driven changes, containment is for user drag.
  - Unselected group, child reflows: Previously resizeNode skipped the update. Now it passes through, but isSizeChanged
  filters it because autoSize: false groups have explicit CSS size matching the stored size. Same effective behavior.
  - Node shrinks below minSize: Previously resizeNode clamped it. Now updateNodes accepts the DOM size. Correct — the DOM
  already rendered at that size, clamping would create a mismatch.
  - Snapping configured: Previously resizeNode snapped. Now updateNodes accepts the DOM size. Correct — snapping is for
  user-driven drag, not content reflow.
  - Active user resize: isResizing guard suppresses the update. Same behavior as before.

piotrblaszczyk (approved):

PR #683, opened 2026-04-28
Dynamic ports labels test

lukasz-jazwa:


apps/angular-demo/src/app/toolbar/batch-test-toolbar.component.ts
```diff
@@ -0,0 +1,207 @@
...
+  private readonly viewportService = inject(NgDiagramViewportService);
+  private readonly ngDiagramService = inject(NgDiagramService);
+
+  exit = output<void>();
+
+  isDebugMode = false;
+  isDynamicPortType = true;
+  private portCounter = 1;
+
+  resetPortCounter(): void {
```
piotrblaszczyk: It's never called anywhere. Maybe it should be called in `enterBatchTest `? If not, it can be removed

apps/angular-demo/src/app/app.component.ts
```diff
@@ -123,10 +126,33 @@ export class AppComponent {
...
+  batchTestMode = signal(false);
+  private savedModelData: Partial<{ nodes: Node[]; edges: Edge[] }> | null = null;
+
+  enterBatchTest(): void {
+    this.savedModelData = this.modelData();
+    this.batchTestMode.set(true);
+    this.modelData.set(generateDynamicPortsTestModel());
+  }
+
+  exitBatchTest(): void {
```
piotrblaszczyk, line 154: Global state leak when exiting batch mode. It restores the model, but debugMode set via `ngDiagramService.updateConfig({ debugMode: true })` stays active globally.

apps/angular-demo/src/app/toolbar/batch-test-toolbar.component.ts
```diff
@@ -0,0 +1,207 @@
...
+      <span class="separator">|</span>
+      <span class="group-label">Measurement:</span>
+      <button [disabled]="!isDynamicPortType" (click)="resizeAllNodes()">Resize All Nodes</button>
+      <button [disabled]="!isDynamicPortType" (click)="repositionPorts()">Reposition Ports (CSS)</button>
+      <span class="separator">|</span>
+      <button (click)="zoomToFit()">Zoom to Fit</button>
+    </div>
+  `,
+})
+export class BatchTestToolbarComponent {
```
piotrblaszczyk, line 43: Inconsistency with OnPush. `DynamicPortNodeComponent` uses `ChangeDetectionStrategy.OnPush`, but `BatchTestToolbarComponent` no. Also it uses plain fields (`isDebugMode`, `isDynamicPortType`) instead of signals. For consistency with the rest of the codebase, switch to signals

lukasz-jazwa: done

apps/angular-demo/src/app/app.component.ts
lukasz-jazwa: it's ok, it's just covenience button to set debug mode in this toolbar, the same exists in previous one

piotrblaszczyk (approved):

PR #685, opened 2026-04-29
Measurement tracking reworked

lukasz-jazwa:


packages/ng-diagram/projects/ng-diagram/src/core/src/types/transaction.interface.ts
```diff
@@ -42,10 +42,10 @@ export interface TransactionOptions {
  * @internal
  */
 export interface InternalTransactionOptions extends TransactionOptions {
+  /** Discovery window in ms after transaction commit — waits for measurement activity to begin. @default 50 */
+  _measurementDiscoveryWindowTimeout?: number;
```
piotrblaszczyk, line 46: @dafult should be 70 here

piotrblaszczyk (approved):

packages/ng-diagram/projects/ng-diagram/src/core/src/types/transaction.interface.ts
lukasz-jazwa: ok, I'll remove it, it's unnecessary dependency on thing which could change

PR #717, opened 2026-05-21
Touchpad gestures adjustments, zoom/pan fixes

Jacek-Synergy:
- Added smooth panning using touchpad
- Added pinch to zoom
- Fixed Shift+wheel not changing pan direction

### DIFF
--- a/packages/ng-diagram/projects/ng-diagram/src/lib/directives/input-events/zooming/zooming-wheel.directive.ts
+++ b/packages/ng-diagram/projects/ng-diagram/src/lib/directives/input-events/zooming/zooming-wheel.directive.ts
@@ -3,6 +3,9 @@ import { FlowCoreProviderService } from '../../../services/flow-core-provider/fl
 import { InputEventsRouterService } from '../../../services/input-events/input-events-router.service';
 import { WheelInputEvent } from '../../../types';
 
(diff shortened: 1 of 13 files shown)


packages/ng-diagram/projects/ng-diagram/src/lib/directives/input-events/zooming/zooming-wheel.directive.ts
```diff
@@ -35,14 +37,32 @@ export class ZoomingWheelDirective {
...
+      return true;
+    }
+
     const flowCore = this.flowCoreProvider.provide();
     const modifiers = this.inputEventsRouterService.getBaseEvent(event).modifiers;
+    return flowCore.shortcutManager.matchesAction('zoom', { modifiers });
+  }
 
-    return !event.zoomingHandled && flowCore.shortcutManager.matchesAction('zoom', { modifiers });
+  private isPinchGesture(event: WheelInputEvent): boolean {
```
lukasz-jazwa, line 66: Can we add comment here explaining the gesture ? Pinch resembles pinching with 2 fingers and in the body of the function we use ctrl modifier (what about mac and cmd ?) which is for mouse wheel zooming gesture

lukasz-jazwa (approved):

PR #759, opened 2026-07-06
Allow no ports in default node

kbcki:


apps/docs/src/content/docs/guides/nodes/nodes.mdx
```diff
@@ -56,10 +56,12 @@ import RotatableResizableNode from '@components/angular/nodes/nodes/rotatable-re
 
 ### Ports
 
-Nodes can expose one or more **ports**, which define the connection points for edges. By convention, the default node includes a single input and a single output port, but you can customize the number, placement, and appearance of ports in your own node templates.
+Nodes can expose one or more **ports**, which define the connection points for edges.
+
+By convention, the default node includes a single input and a single output port, you can hide the ports using global with [`hidePorts: true`](/docs/api/types/configuration/features/defaultnodetemplateconfig/#hideports). If you want to customize the number, placement, and appearance of ports, you need to create your own node templates.
```
lukasz-jazwa: global configuration

lukasz-jazwa:
Add entry to README.md on this change, run `api:update` script and build the solution

apps/docs/src/content/docs/guides/nodes/nodes.mdx
```diff
@@ -56,10 +56,12 @@ import RotatableResizableNode from '@components/angular/nodes/nodes/rotatable-re
...
 
 :::note
-Nodes can also have no ports at all, but without ports, they cannot be connected to other nodes with edges.
+Nodes can have no ports at all. The existing edges will be rendered as [Floating edges](/docs/guides/edges/floating-edges/). The users won't be able to manuall draw edges between nodes.
```
lukasz-jazwa: We need to be more precise here: existing edges will be rendered as floating edges only if they specify `source` and `target` in their data, without having `sourcePort` and `targetPort` provided. When they are provided, and ports are hidden the edge is frozen.

lukasz-jazwa:
I'm starting to doubt that `hideDefaultPorts` and `hidePorts` are precise names - they can introduce  ambiguity - user may expect that ports are still there but invisible. I think that `removeDefaultPorts` and `removePorts` would be easier to understand

lukasz-jazwa (approved):

lukasz-jazwa (approved):

PR #769, opened 2026-07-27
Awaitable service methods

lukasz-jazwa:
# Awaitable state changes

## The problem

When application code changes the diagram — adds nodes, updates data, deletes, pastes — the
change travels through a pipeline (validation middlewares, computed properties, DOM
measurement) before it lands in the model. Until now there was no reliable way to know _when_
that happened. Service methods returned nothing to wait on, so code as simple as:

```ts
this.modelService.deleteNodes(ids);
this.viewportService.zoomToFit(); // operates on stale state — fits the deleted nodes too
```

silently worked on stale state, and teams compensated with polling, timeouts, or wrapping
every call in a measured transaction.

Worse, several failure modes were completely silent: an error thrown inside a middleware froze
every subsequent update with no signal at all, and two transactions overlapping in one task
could silently drop one of them wholesale.

## The change, in one sentence

Every state-changing call now returns a promise that resolves exactly when the change is
committed to the model — `await` it and the very next line reads the updated state — and
everything that used to fail silently now fails loudly.

## What that breaks down into

1. **Awaitable service methods** — every mutating method on the six public services returns
   `Promise<void>` that settles after the full update pipeline. Awaiting is optional; calls
   whose promise you ignore apply the change all the same.
2. **`waitForMeasurements` directly on methods** — the ten methods whose changes can trigger
   DOM measurement (`addNodes`, `addEdges`, `updateNode(s)`, `updateNodeData`,
   `updateEdge(s)`, `updateEdgeData`, `resizeNode`, `paste`) accept
   `options?: { waitForMeasurements?: boolean }`: the promise then resolves only once real
   dimensions are known, so `zoomToFit()`/`centerOnNode()` can follow immediately.
   Deliberately absent on model-only operations (deletes, grouping, rotation) — nothing gets
   measured there, awaiting the method is already enough
   (ADL: `docs/adl/2026-07-21-wait-for-measurements-method-boundary.md`).
3. **Honest failures** — a throwing middleware or user callback rejects the awaited call
   instead of freezing the diagram; the update lock is released and the diagram keeps
   working; late errors are reported via `console.error` instead of vanishing.
4. **Transaction integrity** — `transaction()` returns the commit promise from every overload
   (the sync-callback one included), and overlapping un-awaited transactions each commit
   their own updates (fixes the reported silent-drop bug) with a console warning recommending
   to await them.
5. **Gesture hardening** — once updates can genuinely suspend mid-gesture, the drag, resize,
   rotate and linking flows needed hardening end to end: nodes no longer shift after the
   pointer is released, fast re-grabs are not broken by the previous gesture's cleanup,
   lifecycle events always come in pairs reporting their own gesture's nodes, and gestures
   survive throwing callbacks and mid-gesture node deletions.

Design record: `docs/adl/2026-07-07-awaitable-command-emits.md`.

## Docs

New "Awaiting Changes" section in the state-management guide, "Per-method option" section in
the transactions guide, updated overview and middleware guide (including the one hard rule
for middleware authors: never `await` a mutating call from inside a middleware — fire it
without awaiting).

### Type level: compatible, with two named exceptions

- Every mutating method changed its return type `void` → `Promise<void>` (and
  `NgDiagramService.transaction`'s sync-callback overload `void` →
  `Promise<TransactionResult>`). This is a widening in return position — existing call sites
  compile and behave unchanged; fire-and-forget usage remains valid.
- The new `options?: { waitForMeasurements?: boolean }` parameters are optional trailing
  parameters — compatible.
- **Exception 1**: a subclass overriding a mutating service method with a `void` return stops
  compiling. Mechanical fix: update the override signature.
- **Exception 2**: typed test doubles pinned to the old `void` signatures need their types
  updated. Both exceptions are compile-time-visible; nothing breaks silently at runtime.

### Behavior level (invisible to the API report)

- **Promises resolve later** — after the middleware chain ran and the change committed
  (previously ~1 microtask, before the state applied). Code depending on resolving _before_
  the commit was relying on a bug; no legitimate pattern of that shape is known.
- **Failed updates are observable** — a rejection surfaces to the caller; for
  fire-and-forget callers a failing update appears as an unhandled promise rejection in the
  console instead of disappearing. Intended, but visible.
- **New console warnings** consumers may see after upgrading: overlapping un-awaited
  transactions, and `waitForMeasurements` passed inside an active transaction.
- **Behavior changes only where data was previously lost**: overlapping transactions all
  commit; lifecycle events always pair; gestures survive errors and mid-gesture deletions.
- **Signals are unchanged**: `NgDiagramModelService` signals still refresh asynchronously
  after the promise resolves (synchronous signal freshness is a separate branch).

### Middleware authors

- Never `await` a mutating emit/service call from inside a middleware — the update pipeline
  is not re-entrant and the await deadlocks (such code only appeared to work before because
  promises resolved early). Fire-and-forget calls queue safely.
- A middleware that throws (sync or async) rejects the in-flight update instead of freezing
  every subsequent update.


packages/ng-diagram/projects/ng-diagram/src/core/src/batch-processor/batch-processor.ts
lukasz-jazwa, line 1: **Why this file changed**

`BatchProcessor` batches port/label measurement writes. Its flush callbacks call into the
update pipeline. Before this PR those calls resolved almost immediately, so a flush finished
right away and two flushes could not overlap. Now the calls really wait for the full update
pass, so a flush can stay suspended in the middle (async middleware, busy update lock). That
exposed three hidden problems, fixed here:

1. **Flushes could overlap.** A new flush could start while the previous one was still
   suspended. A delete from flush N could then reach the model _after_ an add from flush N+1
   for the same port/label — stale or missing ports, most likely under virtualization.
   Flushes are now chained on `lastFlush`: a flush starts only after the previous one
   finished.
2. **One failing callback threw away the work of the others.** The pending queues are emptied
   _before_ the callbacks run. So when one callback rejected (rejections are visible now —
   before this PR they were silently ignored), the items collected for the other callbacks
   were lost. Each callback now has its own try/catch.
3. **One failing flush stopped all future flushes.** `flushScheduled` was reset only in the
   middle of `flush()`. An exception thrown before that point left it `true` forever — no new
   flush could ever be scheduled, and every later measurement was silently dropped. The reset
   now happens at the top of `flush()` (the code up to the queue drain is synchronous, so the
   scheduling behavior is the same).

packages/ng-diagram/projects/ng-diagram/src/core/src/command-handler/commands/linking/linking-gesture.ts
lukasz-jazwa, line 1: **Why these files changed** (finish-linking, finish-linking-to-position, start-linking, start-linking-from-position, linking-gesture.ts)

Linking needs the same rule as the other gestures — "clean up only your own gesture" — but
object identity cannot give it here: the linking state object is _replaced_ on every pointer
move, and also by the edges-routing middleware during finishLinking's own pass. So linking
uses a stamp instead: `createLinkingState` (new `linking-gesture.ts`) marks every new gesture
with a growing `_gestureId`, spread copies keep the stamp, and both finish commands clear the
state in `finally` — only when the live stamp still equals the one read at the start. This survives throwing user callbacks
(`validateConnection`/`finalEdgeDataBuilder`/`computeEdgeId` run inside these commands).
`finish-linking-to-position` had no cleanup-on-error at all before.

packages/ng-diagram/projects/ng-diagram/src/core/src/command-handler/commands/move-nodes-lifecycle.ts
lukasz-jazwa, line 1: **Why these files changed** (move-nodes, resize-node, rotate-node)

The lifecycle commands now accept the gesture's node id(s) and pass them on as
`FlowStateUpdate.gestureNodeIds` — the update pass carries its own copy. Without this, the
lifecycle emitters read the _live_ action state at pass time. Under a fast re-grab that
state already belongs to the next gesture, so the previous gesture's end event was lost (the
new gesture starts with empty node ids) or reported the wrong node.

packages/ng-diagram/projects/ng-diagram/src/core/src/input-events/handlers/pointer-move-selection/pointer-move-selection.handler.ts
lukasz-jazwa, line 1: **Why this file changed**

The drag handler is the place where the new timing hurts the most: with an async middleware
in the chain, pointermove events overlap while previous ones are suspended. Before this PR
the handler computed movement from instance fields that changed during those suspensions —
movement was lost or applied twice, and a fast re-drag had its fresh state removed by the
previous drag's late cleanup. The rework: all per-gesture input state lives in one private
`DragGesture` object, created new for each gesture, so **object identity** answers "is this
still my gesture?" after every `await`; `ended` is set synchronously at pointerup, so a
suspended move cannot move nodes after the mouse was released; cleanup runs in `finally`
(awaited emits can reject now — for example a user grouping callback that throws during the
drop) and is guarded by identity, so it never removes a newer gesture's state; and
`moveNodesStart/Stop` carry the gesture's node ids so lifecycle events report the right
nodes (see the lifecycle commands).

packages/ng-diagram/projects/ng-diagram/src/core/src/middleware-manager/middlewares/event-emitter/emitters/node-drag-lifecycle.emitter.ts
lukasz-jazwa, line 1: **Why these files changed** (move-nodes, resize-node, rotate-node)


The lifecycle commands now accept the gesture's node id(s) and pass them on as
`FlowStateUpdate.gestureNodeIds` — the update pass carries its own copy. Without this, the
lifecycle emitters read the _live_ action state at pass time. Under a fast re-grab that
state already belongs to the next gesture, so the previous gesture's end event was lost (the
new gesture starts with empty node ids) or reported the wrong node.

packages/ng-diagram/projects/ng-diagram/src/core/src/middleware-manager/middleware-executor.ts
lukasz-jazwa, line 1: **Why this file changed**

Now that callers really await update passes, pass failure becomes part of the public
contract — and the old executor had no plan for it: an uncaught middleware error left the
pass suspended forever **while holding the update lock**. The diagram froze silently. The
changes: any middleware error rejects the whole pass and releases the lock; the promise
returned by `next()` has a built-in no-op rejection handler, so sync middlewares that call
`next()` without awaiting do not produce "unhandled rejection" noise when a pass fails
(middlewares that `await next()` still receive the real error); a middleware that wakes up
after its pass already ended no longer runs the rest of the chain against that dead pass
(that would consume a live pass's measurement tracking and emit wrong events); and errors
thrown after the pass ended are logged with `console.error` instead of disappearing. The
"end the pass exactly once" bookkeeping moved to its own class — see `pass-settlement.ts`.

packages/ng-diagram/projects/ng-diagram/src/core/src/middleware-manager/pass-settlement.ts
lukasz-jazwa, line 1: **Why this file exists**

Extracted from the executor: one class that owns how a pass ends — the outcome, the
middlewares suspended on `await next()`, and the pass promise callbacks. A pass ends exactly
once, resume order depends on the outcome, late failures are logged, and `lateNextResult()`
answers middlewares that wake up after the pass already ended. This logic used to be written
by hand inside `resolveMiddlewares` and was the source of subtle ordering bugs

packages/ng-diagram/projects/ng-diagram/src/core/src/transaction-manager/transaction-manager.ts
lukasz-jazwa, line 1: **Why this file changed**

The transaction stack removed finished transactions by position (`pop()`). That is only
correct when transactions finish in reverse order of starting — which was true while promises
resolved instantly, and false once they really wait. When a parent transaction finished while
an un-awaited nested transaction was still running, `pop()` removed the nested one instead,
and its queued updates disappeared without any error (the reported bug: two overlapping
`transaction()` calls in one task, the second one fully lost). A finished transaction is now
removed **by identity**. A nested transaction whose parent already finished commits its
updates on its own, with a console warning, instead of merging them into the dead parent.

packages/ng-diagram/projects/ng-diagram/src/core/src/updater/init-updater/init-updater.ts
lukasz-jazwa, line 1: **Why this file changed**

The init completion callback awaits the `init` command pass — which can now really reject
(for example, a custom `ModelAdapter` that throws). Unhandled, that rejection kept the
updater in init mode forever, and every future measurement was ignored. `finish()` now
catches the error, logs it, and completes initialization anyway.

packages/ng-diagram/projects/ng-diagram/src/core/src/flow-core.ts
lukasz-jazwa, line 1: **Why this file changed**

Three fixes for races that became real now that updates can stay suspended much longer:

1. **A root transaction could commit into the wrong transaction.** Commit used to go through
   `applyUpdate`, which routes updates into "the currently active transaction". By the time a
   suspended root transaction commits, a different, unrelated transaction may be active. The
   commit now takes a direct path (`applyUpdateToModel`) that skips that routing.
2. **`waitForMeasurements` tracking could be stolen.** The tracking request was stored
   globally and used by whichever update pass ran next — a concurrent unrelated update could
   take it and register the wrong elements. The request is now stored inside the
   transaction's own pass, under the update lock.
3. **A cancelled or failing pass could break the next one** — it now clears its stored
   tracking request instead of leaving it behind for an unrelated pass.

packages/ng-diagram/projects/ng-diagram/src/lib/public-services/emit-with-measurement-option.ts
lukasz-jazwa, line 1: **Why this file exists**

The single shared implementation behind every `waitForMeasurements`-capable method: the
warn-and-fallback inside an active transaction, the named transaction, and the options
spread that forwards internal tuning fields. Taking `flowCore` as a function argument makes
the synchronous throw on an uninitialized diagram structural — the getter runs at the call
site, so fire-and-forget callers get a catchable synchronous error instead of an unhandled
rejection. Module-private on purpose: not exported through `public-api.ts`, absent from the
API report.

kbcki (approved):

piotrblaszczyk (approved):

PR #779, opened 2026-08-07
Ngd 292 lost final gesture frame

lukasz-jazwa:


packages/ng-diagram/projects/ng-diagram/src/core/src/input-events/handlers/pointer-move-selection/pointer-move-selection.handler.ts
```diff
@@ -212,6 +217,18 @@ export class PointerMoveSelectionEventHandler extends EventHandler<PointerMoveSe
     return true;
   }
 
+  /** Movement vector from `from` to `to` (flow coordinates). */
+  private pointerDelta(from: Point, to: Point): Point {
```
Jacek-Synergy, line 221: This could be moved to NgDiagramMath

Jacek-Synergy (approved):

PR #795, opened 2026-08-21
Add tabbable input to control the diagram's Tab order

lukasz-jazwa:
The diagram host and the watermark link were two Tab stops an application could not remove. `<ng-diagram [tabbable]="false">` takes both out of the sequential Tab order; both stay clickable, stay in the accessibility tree, and keyboard shortcuts keep working once focus is inside the diagram. Defaults to true, so current behavior is unchanged.

The input lives on NgDiagramComponent rather than on the KeyboardInputsDirective host directive: api-extractor's d.ts rollup drops the host-directive reference, so a forwarded input does not compile for consumers installing from npm.

## Backward compatibility

- Default is `true`, so applications that do not use the input keep the current behavior — the diagram container and the watermark link stay in the Tab order.
- **One behavior change:** a static `tabindex` attribute written on `<ng-diagram>` in an application template no longer takes effect. Until now the template attribute won over the directive's static host `tabindex="0"`, so `tabindex="-1"` worked as a workaround to remove the Tab stop. The attribute is now driven by a host binding, and a host binding overwrites the template attribute. Migration is one attribute: replace `tabindex="-1"` with `[tabbable]="false"` (called out in the CHANGELOG).
- A template binding `[attr.tabindex]` on `<ng-diagram>` was never reliable — it always fought the attribute set by the library — and still is not.
- A positive `tabindex` (for example `tabindex="3"` to give the diagram a fixed place in the page's Tab order) has no replacement: `tabbable` produces only `0` or `-1`. Accepted trade-off — positive tabindex values are a WCAG anti-pattern.
- Everything else is additive: one new optional input, no changes to any existing signature or type (the api-report diff is the new input only).


apps/docs/scripts/sync-changelog.js
lukasz-jazwa, line 1: Fix for sync on Windows where sometimes double "Changelog" title existed after sync

Jacek-Synergy (approved):

PR #799, opened 2026-08-31
Hidden elements

Jacek-Synergy:


packages/ng-diagram/projects/ng-diagram/src/lib/components/port/ng-diagram-port.component.ts
```diff
@@ -152,6 +168,15 @@ export class NgDiagramPortComponent extends NodeContextGuardBase implements OnIn
...
       }
     });
+
+    effect(() => {
+      const hidden = this.hidden();
+      const nodeData = untracked(() => this.nodeData());
+      if (!this.isInitialized() || !nodeData) return;
+      // Runtime toggles; the initial value is written in ngOnInit BEFORE the
+      // port registers for measurement.
+      this.flowCoreProvider.provide().templateVisibilityRegistry?.setPortHidden(nodeData.id, this.id(), hidden);
```
lukasz-jazwa, line 184: On Angular 18 this line can crash initialization with NG0600. The registry write is not wrapped in `untracked`, and it can synchronously complete init: `setPortHidden` → `refreshHiddenEntities()` → `tryFinish()` → `finish()` → `setState()`, which writes model signals — and writing signals inside an effect throws NG0600 on Angular 18 (allowed only since Angular 19, so tests on 19.x never see it, but peer deps allow `>=18`).

Trigger: the `hidden` input changes after `ngOnInit` but before init completes, and that change is the last thing init was waiting for.

Fix: wrap the registry write in `untracked(...)` — `NgDiagramHiddenDirective` already does exactly that. The same applies to the label effect in `base-edge-label.component.ts` (line 138).

This constraint could also be recorded in the ADR entry: anything writing to the visibility registry from a reactive context must use `untracked`, because the write can synchronously reach `setState` during init.

CHANGELOG.md
```diff
@@ -9,10 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **Copying a group now includes its content** — `copy`/`cut` cascade to all descendants of the copied nodes and to the edges connecting copied nodes, so pasting a group reproduces its children and internal wiring (previously a pasted group was empty). Cutting a collapsed group no longer destroys its hidden children — they travel through the clipboard with it. Effectively hidden _selected_ elements are skipped by copy (consistent with `deleteSelection`), and pasted hidden content is not selected, so pasting cannot create an invisible selection; the paste-at-cursor position is computed from the visible pasted content only (NGD-101)
```
lukasz-jazwa: The four new entries end with ClickUp IDs (`NGD-101`, `NGD-255`–`NGD-258`), but every other entry in this file uses a GitHub PR link (for example `[#793](...)`). Changelog readers can't open ClickUp — please replace the IDs with PR links.

apps/e2e/tests/hidden.spec.ts
```diff
@@ -0,0 +1,233 @@
+import type { Model } from 'ng-diagram';
+import { expect, test } from './fixtures/diagram';
+
+/**
+ * Hidden elements (NGD-101): a model-level `hidden` flag on nodes and edges.
```
lukasz-jazwa: Repo convention: no ticket numbers in code comments, tests included. Please drop `(NGD-101)` here.

packages/ng-diagram/projects/ng-diagram/src/lib/directives/hidden/ng-diagram-hidden.directive.ts
```diff
@@ -0,0 +1,150 @@
...
+      if (!this.warnedVirtualization) {
+        this.warnedVirtualization = true;
+        console.warn(VIRTUALIZATION_WARNING);
+      }
+      return;
+    }
+
+    const owner = this.resolveOwner();
+    if (!owner) {
+      console.warn(OUTSIDE_TEMPLATE_WARNING);
```
lukasz-jazwa: This warning fires on every toggle, while the virtualization warning above is one-shot (`warnedVirtualization`). If the binding is misplaced and driven by a frequently-changing signal, it will flood the console. Suggest the same one-shot flag here.

packages/ng-diagram/projects/ng-diagram/src/core/src/visibility/template-visibility-registry.ts
```diff
@@ -0,0 +1,129 @@
...
+    this.hiddenPorts.delete(nodeId);
+  }
+
+  /** See {@link removeNodeEntries} — same cleanup for a removed edge. */
+  removeEdgeEntries(edgeId: string): void {
+    this.hiddenEdges.delete(edgeId);
+    this.hiddenLabels.delete(edgeId);
+  }
+
+  clear(): void {
```
lukasz-jazwa: `clear()` has no production caller — only the per-entry cleanup paths are used. Either wire it up (model re-initialization through the same adapter would be the use case) or remove it.

lukasz-jazwa (commented):
Code review notes — details in the inline comments; two repo-level points follow as separate comments (docs changelog sync, removed triage docs).

lukasz-jazwa:
`apps/docs/src/content/docs/changelog.mdx` needs to be regenerated. It's a generated file (`apps/docs/scripts/sync-changelog.js`) that is committed to the repo, and on this branch it's missing all four new changelog entries — it was in sync on the base commit. Please regenerate and commit it together with `CHANGELOG.md`.

The "regenerate together with `CHANGELOG.md`" rule could also be written down in the ADR / contributor docs so it isn't missed next time.

lukasz-jazwa:
The last commit removed `docs/hidden-elements-known-gaps.md` and `docs/hidden-elements-open-bugs.md`, and with them the recorded triage outcomes. What got lost:

- Two won't-fix decisions with their reasoning: unhiding a node during an active drag, and setting `hidden` through a model update during initialization (still waits for the 2 s timeout).
- About ten accepted limitations (e.g. `sendToBack` on a hidden selected node reorders its visible ancestors; `centerOnNode` and `zoomToFit` disagree on hidden policy; a hidden group can keep a sticky `highlightGroup`).
- Two items marked "tracked as a separate task" that have no task yet: (a) `deleteNodes` on a group leaves children with a dangling `groupId`, which makes previously hidden children reappear as orphans; (b) `draggable: false` on a group should freeze its whole subtree.
- NGD-317's description still links to one of the deleted files.

Suggestion: move the accepted decisions and limitations into the ADR entry (and create the two follow-up tasks), so they aren't only recoverable from git history.

lukasz-jazwa:
**`diagramInit` still arrives 2 s late (with console warnings) when ports or edge labels start hidden**

Repro: a port or an edge label hidden from the first render (`[hidden]="true"`). Initialization itself completes fast and the diagram renders correctly, but the `diagramInit` event waits out a 2-second safety timeout and logs `[DiagramInitEmitter] Measurement timeout reached…` with the hidden ports/labels listed as unmeasured.

Cause in one line: `InitUpdater` was taught about the visibility registry, but `DiagramInitEmitter` (`event-emitter/emitters/diagram-init.emitter.ts`) was not — it tracks measurements independently.

Since "no 2 s timeout warning when all visible elements are measured" is part of the acceptance criteria, this belongs in this PR rather than a follow-up. Details and the fix direction are in a comment on the ClickUp task.

lukasz-jazwa:
**A node whose ports start hidden never gets `measuredBounds` — `zoomToFit`, `computePartsBounds` and the minimap frame exclude it entirely**

Repro: a node whose ports are hidden from the first render (`[hidden]="true"`); `zoomToFit` cuts the node off (visible in the demo after deleting the far node). Toggling the ports visible once fixes it permanently; moving the node does not.

Cause in one line: hidden ports never measure, so `isNodeFullyMeasured()` stays false forever and the `measured-bounds` middleware never assigns `measuredBounds` — the node contributes nothing to any bounds computation. Same class as the `DiagramInitEmitter` finding above (a measurement-completeness consumer unaware of the visibility registry), and best fixed together with it.

apps/docs/src/content/docs/guides/conditional-visibility.mdx
```diff
@@ -0,0 +1,98 @@
...
+| ----------------------- | ------------------- | ----------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `hidden` flag / input   | Yes, `display:none` | ✅ No init cost, instant re-measure | Expand/collapse, edit modes, filtering — anything toggled at runtime                                                                                                                         |
+| `@if` (unmounting)      | No                  | ✅ Never registers for measurement  | Content that should genuinely not exist. Note: an unmounted port disappears from `measuredPorts`, so attached edges lose their anchor and fall back to the node                              |
+| raw CSS `display: none` | Yes                 | ⚠️ Discouraged                      | Avoid. Initialization cannot tell "hidden" from "not measured yet", so every load pays the measurement timeout (2 s) before finishing. The zero-size guard only prevents geometry corruption |
```
lukasz-jazwa: Let's remove "The zero-size guard only prevents geometry corruption" - it's internal implementation info not necessary for end user

apps/docs/src/content/docs/guides/conditional-visibility.mdx
```diff
@@ -0,0 +1,98 @@
+---
```
lukasz-jazwa, line 1: Let's add sections about visibility to articles about node, edge, port and label linking to this article

packages/ng-diagram/projects/ng-diagram/src/lib/components/port/ng-diagram-port.component.ts
```diff
@@ -98,6 +98,22 @@ export class NgDiagramPortComponent extends NodeContextGuardBase implements OnIn
...
+   * and is excluded as a linking target and port-snap candidate. Unhiding
+   * re-measures it through the existing ResizeObserver path.
+   *
+   * Edges attached to a hidden port keep the port's last measured geometry
+   * as their anchor; hide the edge itself via its `hidden` flag if it should
+   * disappear with the port.
+   *
+   * @since 1.4.0
+   */
+  hidden = input<boolean>(false);
```
lukasz-jazwa: `hidden` shadows the native global HTML attribute, and without a transform the static form silently means the opposite of native HTML: `<ng-diagram-port hidden />` binds the literal `''`, which is falsy — the port stays **visible**.

Suggestion: `hidden = input(false, { transform: booleanAttribute });` — the static attribute then means "hidden", and bound expressions keep working (Angular-standard coercion, same as Material uses everywhere).

The same one-liner applies to the label input (`base-edge-label.component.ts:75`) and to the directive (`ng-diagram-hidden.directive.ts:50`, `input.required` supports transforms): today a bare `<div ngDiagramHidden>` also binds `''` and silently does nothing, with the transform it means "hidden".

Since 1.4.0 is unreleased, this is free now and a behavior change later.

lukasz-jazwa:
**Suggestion: point the measurement-timeout warnings at the supported hiding mechanisms**

Both timeout warnings (`InitUpdater` and `DiagramInitEmitter`) already guess the cause ("may not be measurable (e.g., display: none)") but don't name the remedy. Since users hitting them are usually hiding content with raw CSS, add one sentence:

> If this content is hidden on purpose, use the `hidden` model flag or `[ngDiagramHidden]` instead of CSS — see the Conditional Visibility guide.

Symptom-triggered, so it fires only when someone actually pays the timeout — no nagging for legitimate native `hidden` on inner template elements. Folds naturally into the `DiagramInitEmitter` fix, which touches the same warning.

apps/docs/src/content/docs/guides/conditional-visibility.mdx
```diff
@@ -0,0 +1,98 @@
...
+title: Conditional Visibility
+description: How to hide nodes, edges, ports and labels — the hidden flag, template inputs, @if, and why raw CSS display none is discouraged
+version: 'since v1.4.0'
+sidebar:
+  badge: New
+---
+
+ngDiagram treats hidden content as a first-class concept. Nodes and edges can be hidden through the model, every template can hide its element declaratively, and ports and edge labels have their own `hidden` inputs — all without blocking initialization or corrupting measurements.
+
+## Key Concepts
```
lukasz-jazwa: **Suggestion: reorder the guide — task first, internals below.**

Most readers land on this page in task mode ("how do I hide a node") and the answer is a 6-line snippet — currently below two sections of concepts. Proposed order:

1. Intro (unchanged — it already summarizes well)
2. Hiding nodes and edges from the model
3. Hiding from templates
4. Hiding ports and edge labels
5. Choosing a pattern (natural close of the "how" half — the reader just met all the patterns)
6. Key Concepts (`hidden` vs `computedHidden` + What hiding does)
7. Virtualization / Clipboard / Persistence — as they are

The move is safe content-wise: the how-to sections don't depend on the concepts section (users only *set* `hidden` there; `computedHidden` matters when *reading* state, which is exactly the internals half).

Two optional touches while at it:
- Rename "Key Concepts" → "How it works" — a "Key Concepts" heading below the content reads oddly after the move.
- Pull the upgrade caution (app-level `hidden: true` data collision) up under "Hiding nodes and edges from the model" — it matters at adoption time, and a task-mode reader copying that snippet never reaches the bottom of the page.

packages/ng-diagram/projects/ng-diagram/src/core/src/middleware-manager/middlewares/hidden-computation/hidden-computation.ts
```diff
@@ -0,0 +1,134 @@
...
+import { computeHiddenNodeIds, isEdgeEffectivelyHidden } from '../../../visibility/effective-visibility';
+import type { TemplateVisibilityRegistry } from '../../../visibility/template-visibility-registry';
+
+/** Node property changes that can alter effective visibility. */
+const HIDDEN_RELEVANT_NODE_PROPS = ['hidden', 'groupId'];
+/** Edge property changes that can alter effective visibility. */
+const HIDDEN_RELEVANT_EDGE_PROPS = ['hidden', 'source', 'target'];
+
+export interface HiddenComputationOptions {
+  /** Distinct middleware name for the pre-pass and finalize instances. */
```
lukasz-jazwa: Let's remove this comment

packages/ng-diagram/projects/ng-diagram/src/core/src/middleware-manager/middlewares/hidden-computation/hidden-computation.ts
```diff
@@ -0,0 +1,134 @@
...
+ * The recompute is idempotent and diff-based: when the pre-pass already
+ * stamped everything, the finalize instance emits nothing.
+ *
+ * @internal
+ */
+export const createHiddenComputationMiddleware = (
+  registry: TemplateVisibilityRegistry,
+  { name, cleanupRemovedEntries = false, onVisibilityChanged }: HiddenComputationOptions
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+): Middleware<any> => ({
```
lukasz-jazwa: The `any` (and the eslint-disable) can go: `Middleware<TName>` only parameterizes the literal type of `name`, so the factory can pass it through a generic:

```ts
export interface HiddenComputationOptions<TName extends string = string> {
  name: TName;
  cleanupRemovedEntries?: boolean;
  onVisibilityChanged?: () => void;
}

export const createHiddenComputationMiddleware = <TName extends string>(
  registry: TemplateVisibilityRegistry,
  { name, cleanupRemovedEntries = false, onVisibilityChanged }: HiddenComputationOptions<TName>
): Middleware<TName> => ({ name, execute: /* unchanged */ });
```

Both call sites in `middleware-manager.ts` then infer precise types (`Middleware<'hidden-computation'>`, `Middleware<'hidden-computation-finalize'>`), which assign fine to the `Middleware | null` fields — `name` is covariant and `execute` doesn't use `TName`. No runtime change.

packages/ng-diagram/projects/ng-diagram/src/core/src/flow-core.ts
```diff
@@ -72,6 +73,26 @@ export class FlowCore {
...
+  readonly templateVisibilityRegistry: TemplateVisibilityRegistry;
+
+  /**
+   * Monotonic counter bumped whenever a middleware pass changed some
+   * element's effective visibility (`computedHidden`). Render caches keyed on
+   * things a visibility toggle does not change (element counts, viewport)
+   * compare it as an O(1) invalidation signal.
+   * @internal
+   */
+  private _visibilityVersion = 0;
```
lukasz-jazwa: Non-blocking refactor suggestion. The mechanism itself is right — a generation counter is the standard O(1) invalidation idiom (Angular signals track producer versions the same way), and it beats the alternatives: a dirty flag breaks with a second consumer, a hidden-count cache key aliases (hide A + unhide B in one pass = unchanged count, stale cache), and an EventManager event is deferred past `setState`, while the strategy needs the answer synchronously in `onChange`.

What could be tidier is the placement: `FlowCore` grows a field + getter + method trio just so one middleware can talk to one render strategy, and the signal travels four hops (middleware → factory options callback → `MiddlewareManager` closure → `FlowCore` method → getter read by `VirtualizedRenderStrategy`). The name is also single-purpose — a second cache with the same need would add `_zOrderVersion`, `notifyZOrderChanged()`, …

Cheap shape: extract a tiny generation object,

```ts
export class ChangeGeneration {
  private v = 0;
  bump(): void { this.v++; }
  get version(): number { return this.v; }
}
```

`FlowCore` only owns it (`readonly visibilityGeneration = new ChangeGeneration()`); the middleware factory and the strategy get the reference directly — the method, the getter and the options callback all disappear, and the pattern is reusable by name. Fine to defer until a second consumer shows up (the pass-coalescing follow-up task is a natural moment).

packages/ng-diagram/projects/ng-diagram/src/lib/components/edge-label/base-edge-label/base-edge-label.component.ts
```diff
@@ -59,6 +62,18 @@ export class NgDiagramBaseEdgeLabelComponent implements OnInit, OnDestroy {
    */
   positionOnEdge = input.required<EdgeLabel['positionOnEdge']>();
 
+  /**
+   * Whether the label is hidden. Defaults to false.
+   *
+   * A hidden label stays mounted as `display: none` and creates no
+   * measurement expectation (it never blocks initialization or
+   * `waitForMeasurements`). Unhiding re-measures it through the existing
```
lukasz-jazwa: Let's change the last sentence to something more general for external user - they don't know internal resizeObserver

packages/ng-diagram/projects/ng-diagram/src/lib/components/port/ng-diagram-port.component.ts
```diff
@@ -98,6 +98,22 @@ export class NgDiagramPortComponent extends NodeContextGuardBase implements OnIn
    */
   originPoint = input<OriginPoint>('center');
 
+  /**
+   * Whether the port is hidden. Defaults to false.
+   *
+   * A hidden port stays mounted as `display: none`, creates no measurement
+   * expectation (it never blocks initialization or `waitForMeasurements`),
+   * and is excluded as a linking target and port-snap candidate. Unhiding
+   * re-measures it through the existing ResizeObserver path.
```
lukasz-jazwa: Let's change the last sentence to something more general for external user - they don't know internal resizeObserver

apps/docs/src/content/docs/guides/conditional-visibility.mdx
```diff
@@ -0,0 +1,102 @@
...
+title: Conditional Visibility
+description: How to hide nodes, edges, ports and labels — the hidden flag, template inputs, @if, and why raw CSS display none is discouraged
+version: 'since v1.4.0'
+sidebar:
+  badge: New
+---
+
+ngDiagram treats hidden content as a first-class concept. Nodes and edges can be hidden through the model, every template can hide its element declaratively, and ports and edge labels have their own `hidden` inputs — all without blocking initialization or corrupting measurements.
+
+## Hiding nodes and edges from the model
```
lukasz-jazwa: Heading wording: in English "hiding X **from** Y" primarily means *concealing X so that Y can't see it* ("hide the file from users"), so "Hiding nodes and edges from the model" reads as hiding them *from* the model — the opposite of what the section does. Same issue in "Hiding from templates" below. A non-native, task-mode reader scanning the headings can pick the wrong section.

Suggestion — name the mechanism, keep the three headings parallel ("what → with what"):

- `## Hiding nodes and edges with the \`hidden\` flag`
- `## Hiding nodes and edges with the \`ngDiagramHidden\` directive`
- `## Hiding ports and edge labels with the \`hidden\` input`

This matches the vocabulary already used in the CHANGELOG ("Declarative `hidden` in templates") and in the directive's TSDoc. Nothing links to these headings by anchor, so the slugs can change freely.

apps/docs/src/content/docs/guides/conditional-visibility.mdx
```diff
@@ -0,0 +1,102 @@
...
+await this.modelService.updateNodes(childIds.map((id) => ({ id, hidden: false })));
+```
+
+Hiding a **group** hides all of its descendants; hiding a **node** hides every edge connected to it. There is no need to touch the edges yourself.
+
+:::caution
+Upgrading from a version before 1.4.0: the library now owns the `hidden` property name on nodes and edges. If your application already stored `hidden: true` on elements for its own purposes, those elements will disappear from the diagram after the upgrade.
+:::
+
+## Hiding from templates
```
lukasz-jazwa: Same "from" ambiguity as the heading above — this reads as hiding something *from* the templates. Suggested: `## Hiding nodes and edges with the \`ngDiagramHidden\` directive` (or, if you prefer to keep it short, "Hiding from inside a template" — "from inside" removes the ambiguity).

lukasz-jazwa (approved):

PR #806, opened 2026-09-14
Delete nodes cascade to descendants

Jacek-Synergy:
## What

`NgDiagramModelService.deleteNodes` now cascades to all descendants of the deleted nodes (children, grandchildren, ...). Edges connected to any deleted node are removed, as before. This matches `deleteSelection`.

## Why

Deleting a group with `deleteNodes([groupId])` left its children in the model with a `groupId` that pointed at a node that no longer existed. A missing parent counts as visible, so children of a hidden group popped back into view as orphans.

The task allowed two fixes: clear the children's `groupId`, or cascade like `deleteSelection`. Cascade was chosen because both delete paths now behave the same and the docs need no special case.

## Backward compatibility

- Signature unchanged. No api-report change.
- Behavior change: `deleteNodes` on a node with descendants now removes the descendants and their edges too. Code that deleted a group and then re-parented the surviving children itself will now find those children gone. To keep the children, call `NgDiagramGroupsService.removeFromGroup` first, then `deleteNodes`.
- Nodes without descendants: no change.


CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`deleteNodes` on a group deletes its whole subtree** — `NgDiagramModelService.deleteNodes` now cascades to all descendants of the deleted nodes, matching `deleteSelection`. Previously the group's children survived with a dangling `groupId` pointing at the deleted parent, and children of a hidden group popped back into view as orphans
```
lukasz-jazwa: Let's add PR url

packages/ng-diagram/projects/ng-diagram/src/core/src/command-handler/commands/__tests__/delete-nodes-cascade.integration.test.ts
```diff
@@ -0,0 +1,70 @@
+/**
+ * Integration test for NGD-320: deleteNodes([groupId]) must cascade to the
```
lukasz-jazwa: Let's remove information about task number

lukasz-jazwa (approved):

